PR #1, opened 2016-07-07
Add cache support for SFDC sessions

pmac:
Not quite ready for merge yet. I'd like to add some tests. Also it seems that the 2 hour limit is rolling, so in prod now it seems that mostly the only times sessions are refreshed are on deploy. I think it'd be best to refresh more often, but still thinking on it.

### DIFF
--- a/news/backends/sfdc.py
+++ b/news/backends/sfdc.py
@@ -1,7 +1,12 @@
 """
 API Client Library for Salesforce.com (SFDC)
 """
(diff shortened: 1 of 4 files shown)


pmac:
May try to get this in soon. There is a SFDC instance refresh tonight, which means that it'll have to use a new instance url at around 1am eastern. This patch will allow it to switch instance URLs, but current prod won't.

PR #5, opened 2016-07-28
Fix #4: Add rate limiting for user update

### LINKED ISSUE #4: Rate limit user update
We added rate limiting of subscriptions in [PR 206](https://github.com/mozilla/basket/pull/206). We've now seen a flurry of `RetryError`s reaching max retries due to seemingly over 1000 of them being for the same token and set of newsletters. My only theories on this so far are that we've either 1) got an issue causing retries due to locking to increase in number exponentially, or 2) someone has submitted the same request 1000 times. Based on number 2 being the least unlikely I propose that we add rate limiting on the `user` view, much the same as was added to `subscribe`.

pmac:


pmac:
@glogiotatidis r?

@jgmize is this and #4 good enough documentation? Or were you looking for something more?

news/views.py
```diff
@@ -712,6 +712,13 @@ def update_user_task(request, api_call_type, data=None, optin=False, sync=False)
                           rate=EMAIL_SUBSCRIBE_RATE_LIMIT, increment=True):
             raise Ratelimited()
 
+    if api_call_type == SET and token and data.get('newsletters'):
+        # only rate limit here so we don't rate limit errors.
+        if is_ratelimited(request, group='news.views.user',
```
glogiotatidis: nit: Feels more like a `group='news.views.set'` to follow the previous `group='news.views.subscribe'`

glogiotatidis:
rgtm r+ ✌️

news/views.py
pmac: A fair point. I used `user` because that's the view from which these requests come. But using the type is probably more clear. Thanks.

jgmize:
> @jgmize is this and #4 good enough documentation? Or were you looking for something more?

I originally asked for a link to a quick summary of the issue on the [internal incidents & responses page](https://mana.mozilla.org/wiki/pages/editpage.action?pageId=52265112), but an actual fix is even better, thanks :)

pmac:
I added a row to the page. Thanks.

PR #8, opened 2016-10-21
Fix bug 1308971: Record all source_urls in SFMC

pmac:


pmac:
@glogiotatidis r?

pmac:
Thanks!

PR #9, opened 2016-11-15
Bug 1302820: Add a subscribe endpoint for use with frontent code

pmac:

PR #10, opened 2016-11-22
Process queue of donation records

pmac:
* Add/Update contacts
* Add donation records
* Read said info from an AWS SQS queue fed by donate.m.o
* Add a donate-worker process type to be run in us-west


pmac:
@glogiotatidis r?

I may need to add more tests, but this one has a deadline of the end of the month, so I wanted to get it in before the US holiday. Let me know what you think of the approach. I'm off Wed - Fri this week, so if you could get some notes for me before Monday I'd really appreciate it. This one is much higher priority than #9.

glogiotatidis:
I'll work on this tomorrow, enjoy the 🦃 !

news/management/commands/process_donations_queue.py
```diff
@@ -0,0 +1,62 @@
...
+    snitch_last_timestamp = 0
+    snitch_id = settings.DONATE_SNITCH_ID
+
+    def snitch(self):
+        if not self.snitch_id:
+            return
+
+        time_since = int(time() - self.snitch_last_timestamp)
+        if time_since > self.snitch_delay:
+            requests.post('https://nosnch.in/{}'.format(self.snitch_id))
```
glogiotatidis: Are you throttling to play nice with DMS or is there another reason?

This is a nice idea for [babis](https://github.com/glogiotatidis/babis) btw, thanks!

news/management/commands/process_donations_queue.py
```diff
@@ -0,0 +1,62 @@
...
+                        continue
+
+                    statsd.incr('mofo.donations.message.received')
+                    try:
+                        process_donation.delay(json.loads(msg.body))
+                    except Exception as e:
+                        statsd.incr('mofo.donations.message.error')
+                        print('ERROR:', e, '::', msg.body)
+                        continue
+                    else:
```
glogiotatidis: If you use `else` in case of an exception you'd keep the msg in the queue which it can eventually result in receiving MaxNumberOfMessages all raising exceptions from the queue, thus blocking the processing. Maybe change this to `finally`?

glogiotatidis: Or keep this and move `msg.delete()` to it's own `finally` block

news/tests/test_tasks.py
```diff
@@ -1,9 +1,11 @@
+from copy import deepcopy
 from urllib2 import URLError
 
+from django.core.cache import cache
 from django.test import TestCase
 from django.test.utils import override_settings
 
-from mock import Mock, patch
+from mock import Mock, patch, ANY
```
glogiotatidis: nit: `ANY` before `Mock`

news/tests/test_tasks.py
```diff
@@ -22,8 +25,97 @@
...
+            'service': u'paypal',
+            'transaction_id': u'NLEKFRBED3BQ614797468093.25',
+        },
+    }
+
+    def test_name_splitting(self, sfdc_mock, gud_mock):
+        data = deepcopy(self.donate_data)
+        gud_mock.return_value = None
+        del data['data']['first_name']
+        data['data']['last_name'] = 'Theodore Donald Kerabatsos'
```
glogiotatidis: :)

news/tasks.py
```diff
@@ -627,6 +627,75 @@ def record_source_url(email, source_url, newsletter_id):
...
+            raise RetryTask('User not yet available')
+
+    # add opportunity
+    donation = {
+        'RecordTypeId': settings.DONATE_OPP_RECORD_TYPE,
+        'Name': 'Foundation Donation',
+        'Donation_Contact__c': user_data['id'],
+        'StageName': 'Closed Won',
+        'CloseDate': timestamp,
+        'Amount': float(data['donation_amount']),
```
glogiotatidis: Is there a possibility that you receive `donation_amount` in a non US-locale? E.g. 30 something and 50 cents it's `30,50` in Greek compared to `30.50`

news/management/commands/process_donations_queue.py
pmac: Just playing nice with DMS. We don't need to ping every time as they'll be coming in quite frequently during the height of the campaign. And yes indeed, it would be a great addition to babis! Also thanks for the link to babis, I had not heard of it. A very useful utility!

news/tasks.py
pmac: It should most often be a real float, but can apparently be a string from one of the payment processors, but in testing it always seems to be in the US-locale form. I'll ask the mofo folks. Good catch.

pmac:
I've fixed (I think) the exception handling and asked about the currency value. Should be ready for another r?

news/tasks.py
glogiotatidis: Does this mean that all transactions are translated into dollars by the time they get received by basket?

If I select "EUR" over at donation.m.o and type `2,1` would basket see `2.1`?

glogiotatidis:
r+ from me given that the `amount` locale answer is clear enough.

Thanks @pmac!

news/tasks.py
cadecairos: The values are extracted from the PayPal/Stripe APIs, so we can be sure that they'll always be in the US-locale.

glogiotatidis: KK, thanks for clarifying.

PR #11, opened 2017-03-03
Fix bug 1082122: Add support for specific 5 char langs

pmac:
This is done with a setting so that we can add more in future without code changes.


metadave:
👍

pmac:
@metadave thanks for taking a look! You think this is good enough to merge and test in dev?

metadave:
yup

PR #15, opened 2017-03-30
Switch docker base image to python:2.7-alpine

pmac:
Seems to be the easiest way to get tini (small init system that handles subprocess reaping and signals) as well as the latest Python 2.7.


pmac:
@jgmize @glogiotatidis perhaps you have opinions? Seems to work well in my local testing.

PR #17, opened 2017-04-12
Fix bug 1353161: Import FxA Activity data into SFMC

pmac:
* Download csv files from s3
* Parse csv files and get the most recent timestamps per fxa_id
* Update said timestamps in a Data Extension in SFMC
* Cache the timestamps in Redis to avoid so many SFMC API calls

These files are FxA login timestamps per day. Each one contains around 20M rows. After processing all 8 (max in the bucket at a time) there are around 10M records to update. This will take quite a while per run.

### DIFF
--- a/bin/deis-cli-install.sh
+++ b/bin/deis-cli-install.sh
@@ -1,19 +1,10 @@
 #!/bin/sh
 
 # included in the repo because it's only available over http from deis.io
+# http://deis.io/deis-cli/install.sh
 
 # install current version unless overridden by first command-line argument
-cd ~/docker
-
-VERSION=${1:-1.13.2}
-
-if [[ -x ./deis ]]; then
-  if [[ "$VERSION" == $(./deis version) ]]; then
-    exit 0
-  else
-    rm ./deis
-  fi
-fi
+VERSION=${1:-1.13.3}
 
 # catch errors from here on out
 set -e
@@ -21,7 +12,7 @@ set -e
 # determine from whence to download the installer
 PLATFORM=`uname | tr '[:upper:]' '[:lower:]'`
 DEIS_INSTALLER=${DEIS_INSTALLER:-deis-cli-$VERSION-$PLATFORM-amd64.run}
-DEIS_BASE_URL=${DEIS_BASE_URL:-https://s3-us-west-2.amazonaws.com/get-deis}
+DEIS_BASE_URL=${DEIS_BASE_URL:-https://getdeis.blob.core.windows.net/get-deis}
 INSTALLER_URL=$DEIS_BASE_URL/$DEIS_INSTALLER
 
 # download the installer archive to /tmp
--- a/news/management/commands/process_fxa_data.py
+++ b/news/management/commands/process_fxa_data.py
@@ -1,0 +1,163 @@
+from __future__ import print_function, unicode_literals
+
+from email.utils import formatdate
+
+from django.conf import settings
+from django.core.cache import cache
+from django.core.management import BaseCommand, CommandError
+
+import babis
+import boto3
+from apscheduler.schedulers.blocking import BlockingScheduler
+from django_statsd.clients import statsd
+from pathlib2 import Path
+from pytz import utc
+from raven.contrib.django.raven_compat.models import client as sentry_client
+
+from news.backends.sfmc import sfmc, NewsletterException
+
+
+TMP = Path('/tmp')
+BUCKET_DIR = 'fxa-last-active-timestamp/data'
+DATA_PATH = TMP.joinpath(BUCKET_DIR)
+FXA_IDS = {}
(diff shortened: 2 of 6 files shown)


pmac:
@glogiotatidis got any time to look at this one soon? The email team wants this in testing asap.

glogiotatidis:
> @glogiotatidis got any time to look at this one soon? The email team wants this in testing asap.

Sorry for the delay I's on PTO. I'll work on this today.

news/management/commands/process_fxa_data.py
```diff
@@ -0,0 +1,154 @@
...
+            else:
+                set_file_done(tmp_path)
+
+            # done with file either way
+            tmp_path.unlink()
+
+    return all_fxa_times
+
+
+@schedule.scheduled_job('interval', id='process_fxa_data', days=1)
```
glogiotatidis: Add `max_instances=1` to ensure that long runs don't conflicts? 

https://apscheduler.readthedocs.io/en/latest/userguide.html#limiting-the-number-of-concurrently-executing-instances-of-a-job

glogiotatidis: Also `days=1` means that it will run in 24 hours from the moment you started the app. If the job runs in the monirng and then you do a app push (and hence restart) in the evening (+6 hours) the job will run in 24+6 = 30 hours after the last run.

This may be a problem or not depending on the requirements, I'm just bringing this here because i've been beaten by missing dead man's snitches.

Alternatively you can specific the exact time to run ala cron 
`scheduled_job('cron', month='*', day='*', hour='3', minute='10')`

news/management/commands/process_fxa_data.py
```diff
@@ -0,0 +1,154 @@
...
+
+def set_fxa_time(fxa_id, fxa_time):
+    try:
+        sfmc.upsert_row(settings.FXA_SFMC_DE, {
+            'FXA_ID': fxa_id,
+            'Timestamp': formatdate(timeval=fxa_time, usegmt=True),
+        })
+    except NewsletterException:
+        sentry_client.captureException()
+        # try again later
```
glogiotatidis: Later would be the next time this `fxa_id` appears in a new logfile in s3? 

Could we make this a celery async task and automatically retry a couple of times or is this over-engineered for the type of data we have?

news/management/commands/process_fxa_data.py
```diff
@@ -0,0 +1,154 @@
...
+        if not tmp_path.name.endswith('.csv'):
+            continue
+
+        if file_is_done(tmp_path):
+            continue
+
+        if not tmp_path.exists():
+            print('getting ' + obj.key)
+            print('size is %s' % obj.size)
+            tmp_path.parent.mkdir(parents=True, exist_ok=True)
```
glogiotatidis: `.parent.` got me thinking about switching to pathlib. 👍

news/management/commands/process_fxa_data.py
```diff
@@ -0,0 +1,154 @@
...
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        if not all(getattr(settings, name) for name in ['FXA_ACCESS_KEY_ID',
+                                                        'FXA_SECRET_ACCESS_KEY',
+                                                        'FXA_S3_BUCKET']):
+            raise CommandError('FXA S3 Bucket access not configured')
+
+        main()
+        schedule.start()
```
glogiotatidis: Maybe useful to activate `schedule.start` via an extra flag to the command, something like  `python manage.py process_fxa_data --daemon`. This way we'll be able to run one-offs, specially useful until we get this running?

bin/deis-cli-install.sh
```diff
@@ -1,27 +1,18 @@
...
+VERSION=${1:-1.13.3}
 
 # catch errors from here on out
 set -e
 
 # determine from whence to download the installer
 PLATFORM=`uname | tr '[:upper:]' '[:lower:]'`
 DEIS_INSTALLER=${DEIS_INSTALLER:-deis-cli-$VERSION-$PLATFORM-amd64.run}
-DEIS_BASE_URL=${DEIS_BASE_URL:-https://s3-us-west-2.amazonaws.com/get-deis}
+DEIS_BASE_URL=${DEIS_BASE_URL:-https://getdeis.blob.core.windows.net/get-deis}
```
pmac: yeah 😕

news/management/commands/process_fxa_data.py
pmac: I think that would be a bit much, also I don't want to try to shove millions of additional items on the queue per day. I think this will be fine. The idea is that this timestamp won't be saved in the local or redis cache, so when this user appears later it'll be added then. Active users of FxA appear in these files multiple times per file, so if they're active they should show up every run.

PR #18, opened 2017-04-18
Fix missing last name errors

pmac:
Requests come in with blank last name will result in errors.


pmac:
https://sentry.prod.mozaws.net/marketing/basket-prod-us-west/issues/340713/

PR #20, opened 2017-05-16
Fix bug 1355567: Record __NONE__ for empty source_url

pmac:
Also fix bug 1308971 by recording source_url for all subscription requests, and ones from email pref center.


pmac:
Looking good. Tests pass. Let's try it in dev.

PR #23, opened 2017-06-16
Replace request.is_secure() with is_secure(request)

jgmize:
Respect new HTTPS setting & env var


pmac:
We're just removing the code in #26. Thanks though!

PR #27, opened 2017-07-13
Rearrange files to modern Django standards

pmac:
* Create base "basket" package
* Move apps, settings, wsgi, and root urls into said package
* Alter all imports to reflect new layout


pmac:
Tried this before in mozilla/basket#192, but I think things are more calm now and I can get it done.

If tests pass I'm just going to self r+ this as there are no code changes.

PR #28, opened 2017-07-19
Rename mozmar github repo to mozmeao

pmac:


metadave:
👍

pmac:
See also mozmar/meta#1

PR #30, opened 2017-08-04
Process donation events

pmac:
Events like charge failures, refunds, and disputes.

See mozilla/donate.mozilla.org#1765

### DIFF
--- a/basket/news/backends/sfdc.py
+++ b/basket/news/backends/sfdc.py
@@ -322,7 +322,7 @@ class SFDC(object):
     @time_request
     def update(self, record, data):
         """
-        Update data in an existing record.
+        Update data in an existing contact record.
 
         @param record: current contact record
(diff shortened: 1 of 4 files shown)


desiree-mofo:
For all status values noted with "Y" in column B in this worksheet (https://docs.google.com/spreadsheets/d/1A_ZYhjJ10be3uKSQVMXRqKK9Ku6tz22Vt62gL3xclF8/edit#gid=2007576112), please update the following fields in Salesforce:

Opportunity.Stage = "Closed Lost"
Opportunity.PMT_Reason_Lost__c = failure_code (for bank failures) OR reason from disputes or refunds
Opportunity.PMT_Type_Lost__c = event_type

pmac:
I think this is good enough for a first try. Let's give it a go on dev and see how it works.

PR #35, opened 2017-08-14
Bug 1358123: Add handler for FxA deleted queue

pmac:


pmac:
@rfk This expects the message from the queue to be simply

```json
{
  "event": "delete",
  "uid": "<the fxa id>"
}
```

But from the look of the [code you pasted in the bug](https://github.com/mozilla-services/tokenserver/blob/master/tokenserver/scripts/process_account_events.py#L89) that structure might be embedded deeper under a `Message` key. Do I need to alter what I have? We'll find out in testing of course.

rfk:
Yeah, unfortunately due to the way that SQS works, I think you'll get a top-level "Message" key with the actual JSON payload in it as a string.

pmac:
Interesting. I don't get that from other SQS queues. Perhaps it's a consequence of the libraries we're using to access SQS? Not a big deal either way, just want to get it right :) Thanks!

basket/news/tasks.py
```diff
@@ -201,6 +202,39 @@ def gmttime():
...
+        return
+
+    _update_fxa_info(email, lang, fxa_id)
+
+    if subscribe:
+        upsert_user.delay(SUBSCRIBE, {
+            'email': email,
+            'lang': lang,
+            'newsletters': settings.FXA_REGISTER_NEWSLETTER,
+            'source_url': settings.FXA_REGISTER_SOURCE_URL,
```
pmac: @rfk I think this should work well. Do you know if you're currently sending the `source_url` based on the actual FxA domain that sent the event? I'm not sure it matters much for us, but I could do something fancy with this based on whether we're consuming the dev, stage, or prod queue.

basket/news/tasks.py
```diff
@@ -201,6 +202,39 @@ def gmttime():
...
+        })
+
+
+@et_task
 def add_fxa_activity(data):
```
pmac: @rfk after the other things, I believe the only other HTTP requests you send us are for this function which records account activity and includes the FxA ID, the `user_agent`, and a `first_device` boolean. Is this another event that will be on the queue?

basket/news/tasks.py
rfk: It's a config value that we *could* change for each environment, but I think right now it just defaults to "accounts.firefox.com" even in dev.  Let me know if you have a use for us changing it and I'll file a bug about updating the config.

basket/news/tasks.py
rfk: Yes, there will be a "login" event in the queue which we currently handle here:

  https://github.com/mozilla/fxa-basket-proxy/blob/master/lib/events/index.js#L86

Our handling of this event does two things.  It registers the activity by posting to `/fxa-activity/`:

  https://github.com/mozilla/fxa-basket-proxy/blob/master/lib/events/index.js#L114

And it also does a bit of magic where, if the user logged in with certain special values of the "utm_campaign" query parameter, it adds them to special subscriptions in basket:

  https://github.com/mozilla/fxa-basket-proxy/blob/master/lib/events/index.js#L94

It's a lightly hacky way for us to tunnel metrics info through FxA for various first-run-page experiments.

If you're happy to take over handling of those events as well, that'll be great!  We'll have to be careful though because IIUC the `/fxa-activity/` part is not idempotent.

> after the other things, I believe the only other HTTP requests you send us
> are for this function which records account activity

FWIW we also provide a "communication preferences" screen in FxA settings, where the user can change their choice of the marketing opt-in checkbox that they made during signup.  We currently action this by making calls to the existing basket web API, but we could consider whether there's a better way.

pmac: > FWIW we also provide a "communication preferences" screen in FxA settings, where the user can change their choice of the marketing opt-in checkbox that they made during signup. We currently action this by making calls to the existing basket web API, but we could consider whether there's a better way.

That user interaction seems like a fine place for using the basket api, and if you're happy to keep doing it that's fine with me. I think we can move the activity handling to the queue, but that's a good point about it being different and not idempotent... Perhaps we should hold of on that bit and do it in a separate bug and sometime later?

pmac: Filed #37 to cover this BTW.

basket/news/management/commands/process_fxa_queue.py
```diff
@@ -0,0 +1,90 @@
...
+                    try:
+                        FXA_EVENT_TYPES[event_type].delay(event)
+                    except Exception:
+                        # something's wrong with the queue. try again.
+                        statsd.incr('fxa.events.message.queue_error')
+                        sentry_client.captureException(tags={'action': 'retried'})
+                        continue
+
+                    statsd.incr('fxa.events.message.success')
+                    msg.delete()
```
glogiotatidis: Would it be useful to delete the msg just before `et_task` completes to ensure that we don't miss any msgs in case salesforge is down for example?

pmac: I was thinking we could rely on celery retries for those cases, but it is possible to lose some if it's down for a long time. I'm not currently passing the message object to the task (not sure how well it'd go on the celery queue), but I could probably figure out how to delete the message from the task instead of the consumer; I'm just not sure it's worth going to that trouble. What do you think?

The other option is to just not use celery tasks for this and to just record the messages as they come in from the queue consumer process. I didn't do that because celery gives us retries and much more scalability, but it could work better in the case of message deletion timing.

glogiotatidis: I hear you, I don't know if that's worth the trouble either. 

Actually re-reading this made me think: Do you plan to run this command periodically? If yes how will it get triggered?

pmac: No, it just runs indefinitely via that infinite loop. Same as the `process_donations_quqeue` command.

basket/news/tasks.py
```diff
@@ -201,6 +203,45 @@ def gmttime():
...
+    if email_is_testing(email):
+        return
+
+    lang = get_best_language(get_accept_languages(locale))
+    if not lang:
+        return
+
+    _update_fxa_info(email, lang, fxa_id)
+
+    if subscribe:
```
pmac: @rfk I guess we'll need to implement this bit as well, and add those new campaigns from the bug?

https://github.com/mozilla/fxa-basket-proxy/blob/13ed0f5206f9150566135d3a0973ac358cf1c0f7/lib/events/index.js#L96

pmac: Actually nevermind. Won't need to do that until we takeover the "login" events in #37. I'll add a note there to be sure to do this bit.

pmac:
@rfk @jbuck I think we're ready to go with this. Let's coordinate on a time when we can:

1. Get messages flowing to the stage and prod queues
2. Deploy basket changes
3. Stop posting to the `/fxa-register/` basket endpoint

jbuck:
@pmac do we need to do these all around the same period of time, or can I get the messages flowing to stage & prod queues before the basket & fxa deploys?

pmac:
It'd be nice if it were "around" the same time just to reduce the number of duplicate queue messages and POSTs coming from FxA, and to minimize the queue size at basket deploy. Is there something in FxA to deploy? The stopping of POSTing to basket for these things I guess?

But to be clear they don't have to be at exactly the same time, but the closer the better.

jbuck:
It looks like there's a "Purge Queue" action I can do, so I can deploy the queues to stage & prod now and purge before we deploy the FxA changes.

pmac:
> It looks like there's a "Purge Queue" action I can do, so I can deploy the queues to stage & prod now and purge before we deploy the FxA changes.

Works for me. Thanks! 👍

pmac:
Just to be clear, this code will go to dev and stage when I merge to master. I'll tag and push to the `prod` branch when we're ready for production.

rfk:
> Stop posting to the /fxa-register/ basket endpoint

This should be pretty trivial, we can include this change in FxA train-96, going to stage this week and deploying next week.

rfk:
So we could aim to give live week of 25th September on both sides?

pmac:
Sounds good to me. Thanks!

rfk:
> So we could aim to give live week of 25th September on both sides?

FYI this is in stage right now on our side

pmac:
Great. Deploying to stage here as well then.

pmac:
These changes have been deployed to basket dev and stage and the queue processor containers have been started. Let's get some more test data flowing :)

rfk:
I just signed up as "ryan-loves-newsletters@rfk.id.au" and checked the opt-in box, do you see any events from me?

pmac:
Actually I'm having trouble with the stage queue :( 

botocore.exceptions.ClientError: An error occurred (AWS.SimpleQueueService.NonExistentQueue) when calling the ReceiveMessage operation: The specified queue does not exist or you do not have access to it.

I get that with the prod and stage settings. The dev queue works. Not sure why yet.

pmac:
I've had a look around the IAM roles for access to these queues and it looks right to me. Perhaps @jbuck and I can troubleshoot tomorrow.

jbuck:
I've fixed this in stage - it'll be going out with prod later today

pmac:
Confirmed the stage queue is now working! Or at least it's allowing me to connect (haven't verified any actual messages yet). Prod queue is still broken but it sounds like you're on it. Thanks @jbuck ! 🥇

Issue #36, opened 2017-08-31
Expand SMS function to be international


pmac:
We'll be expanding our so called "send to device widget" on bedrock to work with several more countries (currently only works for phone numbers in the US) ([bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1389176)). Basket will need to be updated in the following ways:

* SMS code admin will need to expand to include fields for languages, countries, and description (much like transactional emails)
  * It's possible we could just use the message ID to do this (e.g. `download-firefox-mobile-es-es`), but adding real admin fields feels more reliable and less error prone.
* Phone number validation will need to adjust to include the range of number formats from the new countries
* SMS endpoint will need to accept new parameters "country" and "lang"
* If the message ID is not supported in the language specified it should fall back to the "en" one if it exists

pmac:
Done!

PR #38, opened 2017-09-08
Bug 1389176: fix issue #36: Add i18n to SMS messages

pmac:
* Add support for different countries and languages to basket SMS admin
* Add support for parsing and formatting international phone numbers via the python port of libphonenumber from Google

**NOTE**: This should be deployed in three steps, one for each commit in this PR. So the commits should **NOT** be squashed and should be tagged and deployed separately. This will ensure that the code and data migrations make it to prod in the proper order and not cause errors.


basket/news/views.py
```diff
@@ -464,30 +464,30 @@ def invalid_email_response():
...
-    messages = get_sms_messages()
+    country = request.POST.get('country', 'us')
+    language = request.POST.get('lang', 'en-US')
     msg_name = request.POST.get('msg_name', 'SMS_Android')
-    if msg_name not in messages:
+    vendor_id = get_sms_vendor_id(msg_name, country, language)
+    if not vendor_id:
         return HttpResponseJSON({
             'status': 'error',
             'desc': 'Invalid msg_name',
```
glogiotatidis: Change error message to `invalid msg_name, country, language combination`?

basket/news/models.py
```diff
@@ -308,9 +307,21 @@ class SMSMessage(models.Model):
         max_length=200, blank=True,
         help_text='Optional short description of this message'
     )
+    language = LocaleField(default='en-US')
+    country = CountryField(default='us')
 
     class Meta:
-        verbose_name = "SMS message"
+        unique_together = ['message_id', 'language', 'country']
+        verbose_name = "Localized SMS message"
```
glogiotatidis: nit: `s/"/'/g`

basket/news/newsletters.py
```diff
@@ -8,21 +8,16 @@
...
+from basket.news.models import Newsletter, NewsletterGroup, LocalizedSMSMessage, TransactionalEmailMessage
 
 
 __all__ = ('clear_newsletter_cache', 'get_sms_messages', 'newsletter_field',
            'newsletter_name', 'newsletter_fields')
 
 
 CACHE_KEY = "newsletters_cache_data"
-SMS_CACHE_KEY = "sms_messages_cache_data"
+SMS_CACHE_KEY = "local_sms_messages_cache_data"
```
glogiotatidis: nit: `s/"/'/g`

basket/news/newsletters.py
```diff
@@ -46,15 +41,21 @@ def get_sms_messages():
     data = cache.get(SMS_CACHE_KEY)
     if data is None:
         # TODO have this be an empty dict when SMS_MESSAGES is removed.
```
glogiotatidis: Remove comment?

requirements/base.txt
```diff
@@ -132,3 +132,6 @@ APScheduler==3.3.1 \
     --hash=sha256:f68874dff1bdffcc6ce3adb7840c1e4d162c609a3e3f831351df30b75732767b
 tzlocal==1.3.2 \
     --hash=sha256:07dee68c9f5b9ca665762eb0860418cc289bd561f7e32a4ec37a09426895e7bd
+phonenumbers==8.8.1 \
```
glogiotatidis: nice library 👍

pmac:
@glogiotatidis thanks! I've updated the things you mentioned. I also made a change that was just bugging me. I get the `vendor_id` for the SMS message in the view, but then just drop it and send the message ID, country, and language via the queue. I have to only add optional kwargs to the task since old tasks with the old function signature will be on the queue when it's deployed. So instead I've added `vendor_id` as the optional kwarg and will use that instead of discovering it again in the task. This seems cleaner to me. I'll follow up with more changes to the signature to clean it up.

glogiotatidis:
Perfect! ship it!

PR #46, opened 2017-10-20
Fix #45: Strip name data for donations

### LINKED ISSUE #45: Donation fails if last name is spaces
It tries to process the name as if it's real but then there's nothing:

https://sentry.prod.mozaws.net/operations/basket-prod-us-west/issues/673366/

Just need to strip whitespace.

pmac:


pmac:
Looks good and tests pass.

Issue #42, opened 2017-10-03
Add FxA Oauth support


pmac:
https://bugzilla.mozilla.org/show_bug.cgi?id=1051808

pmac:
PR is merged, so this is done. @rfk let me know when you'd like to test with stage. The code should be deploying there now.

PR #49, opened 2017-10-26
Fix bug 1051808: Add FxA OAuth

pmac:
Support FxA Oauth headers for certain requests.


pmac:
also re #42.

pmac:
@rfk mind taking a look through this as well and make sure I'm doing the FxA dance correctly?

pmac:
Anyone with time to review this one? I guess it's not a rush, but it would be nice to get landed so FxA can make the changes on their end and we can test this.

pmac:
@rfk ping again now that we've sorted notifications :)

basket/news/views.py
```diff
@@ -763,10 +769,10 @@ def update_user_task(request, api_call_type, data=None, optin=False, sync=False)
                 }, 400)
 
             if api_call_type != UNSUBSCRIBE and nl in private_newsletters:
-                if not has_valid_api_key(request):
+                if not is_authorized(request, data.get('email')):
                     return HttpResponseJSON({
                         'status': 'error',
-                        'desc': 'private newsletter subscription requires a valid API key',
+                        'desc': 'private newsletter subscription requires a valid API key or Oauth',
```
rfk: nit: technically this should be capitalized like "OAuth" ;-)

basket/news/utils.py
```diff
@@ -95,14 +100,66 @@ def email_is_blocked(email):
...
+        return False
+
+    # Grab the token out of the Authorization header
+    authorization = request.META.get('HTTP_AUTHORIZATION')
+    if not authorization:
+        return False
+
+    authorization = authorization.split(None, 1)
+    if authorization[0].lower() != 'bearer' or len(authorization) != 2:
+        return False
```
rfk: Do you think this header parsing is something we could provide via a more handy helper function in PyFxA itself?

basket/news/utils.py
```diff
@@ -95,14 +100,66 @@ def email_is_blocked(email):
...
+
+    token = authorization[1]
+    oauth = fxa.oauth.Client(server_url=server_urls['oauth'])
```
rfk: By default, this `oauth.Client` instance will keep a little in-memory cache of recently-seen tokens.  I'm not sure that'll be worth it for your use-case here, where you don't really expect to see multiple requests with the same token.  If you like can disable it by passing `cache=None` in the constructor here.

basket/news/utils.py
```diff
@@ -95,14 +100,66 @@ def email_is_blocked(email):
...
+    profile = fxa.profile.Client(server_url=server_urls['profile'])
```
rfk: You could consider making these `Client` instances globals to avoid a small amount of initialization overhead on each request, but I don't think it's too big of a deal.

rfk: Although I guess that could make it harder to test...

basket/news/utils.py
pmac: Wouldn't hurt. It's not completely trivial to correctly handle all the possible values this header might have. I'd certainly have used such a thing :) It'd be nice if you could just pass in a request object to something like that, but there is no one standard python request object and I'm sure you don't want to start having to support any set of web frameworks. But a function to parse the header value and return the token or `None` would be nice.

basket/news/utils.py
pmac: It would, but I can probably get around that. I'll see what I can do.

basket/news/utils.py
pmac: I've looked at the code and the cache has a default TTL of 5 min, so I think it seems like no harm to keep and possibly useful in some situations. Leaving the defaults. Thanks for the tip!

rfk: Right, it expires aggressively, but there's no upper limit on the number of items in the cache if you happen to have many millions of unique accesses in that 5-minute window.

pmac: Hopefully you're not planning on that ;)

PR #59, opened 2018-01-19
Upgrade to Django 1.11.

pmac:
This is the first pass. The tests are passing (for me locally at least). Need more deployment and manual testing before merge.

### DIFF
--- a/basket/news/middleware.py
+++ b/basket/news/middleware.py
@@ -23,11 +23,13 @@ class GraphiteViewHitCountMiddleware(GraphiteRequestTimingMiddleware):
 
 
 class HostnameMiddleware(object):
-    def __init__(self):
+    def __init__(self, get_response):
         values = [getattr(settings, x) for x in ['HOSTNAME', 'DEIS_APP',
                                                  'DEIS_RELEASE', 'DEIS_DOMAIN']]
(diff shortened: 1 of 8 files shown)


basket/news/models.py
```diff
@@ -259,11 +263,11 @@ def notify_stewards(self, name, email, lang, message):
...
         try:
-            stewards = self.localestewards_set.get(locale=lang)
-            emails = stewards.emails
+            stewards = self.stewards.get(locale=lang)
+            emails = stewards.emails_list
         except LocaleStewards.DoesNotExist:
-            emails = self.default_steward_emails
-        emails = emails.split(',')
+            emails = self.default_steward_emails_list
+        print emails
```
glogiotatidis: extraneous `print`

basket/news/middleware.py
```diff
@@ -23,11 +23,13 @@ def process_view(self, request, view_func, view_args, view_kwargs):
 
 
 class HostnameMiddleware(object):
-    def __init__(self):
+    def __init__(self, get_response):
```
glogiotatidis: 👍

basket/news/models.py
pmac: doh! good catch!

PR #60, opened 2018-01-19
Fix bug 1384385: Handle FxA primaryEmailChanged messages

pmac:


pmac:
cc @rfk

basket/news/tasks.py
```diff
@@ -212,6 +212,24 @@ def fxa_source_url(metrics):
     return source_url
 
 
+@et_task
+def fxa_email_changed(data):
+    ts = data['ts']
```
glogiotatidis: q: Is `data['ts']` type `float`?

rfk: Yes, it should be a float in seconds; generated here for reference:

  https://github.com/mozilla/fxa-auth-server/blob/d205c9abee380e250413570b9873561f55bb5068/lib/log.js#L128

pmac:
@rfk so in your opinion is this ready to roll out on my side? are these messages coming through the queue now, or they will be soon? I don't think timing is important on this one right?

rfk:
Yep, should be ready to go with no extra deploy dependencies

pmac:
Fantastic. Many thanks to you both!

PR #63, opened 2018-02-23
Improve healthcheck URL

pmac:
* Add django-allow-cidr
* Add healthz and liveness from kitsune


basket/settings.py
```diff
@@ -133,6 +135,7 @@ def path(*args):
 ]
 
 MIDDLEWARE = (
+    'allow_cidr.middleware.AllowCIDRMiddleware',
     'django.middleware.security.SecurityMiddleware',
     'basket.news.middleware.HostnameMiddleware',
```
glogiotatidis: In https://github.com/mozilla/kitsune/pull/3080 i added `hostnamemiddleware` at the very top. Do you think allowcidrmiddleware should come first?

pmac: I don't think it's a big deal for that one. It's only processing responses, so it could really go anywhere in the list (though Django 1.10+ does care more about response middleware order). `AllowCIDRMiddleware` should go very high in the list because it's to do with security and you don't want much if any code executing before it does the checks. I'd probably leave the CIDR and Security middlewares as 1 and 2, but it really shouldn't matter in this case.

PR #64, opened 2018-02-26
Switch health checks to django-watchman

pmac:
That was fairly painless. I added the HTTPS exception as `r'^/healthz/(ping/)?$'` so that the human readable URL (`/healthz/dashboard/`) would get the HTTPS redirection.

Here's a screenshot of it in action locally:


basket/settings.py
```diff
@@ -178,12 +179,18 @@ def path(*args):
 SECURE_CONTENT_TYPE_NOSNIFF = config('SECURE_CONTENT_TYPE_NOSNIFF', default=True, cast=bool)
 SECURE_SSL_REDIRECT = config('SECURE_SSL_REDIRECT', default=False, cast=bool)
 SECURE_REDIRECT_EXEMPT = [
-    r'^healthz/$',
-    r'^readiness/$',
+    r'^healthz/(ping/)?$',
```
jgmize: Changing the readiness URL would mean temporarily removing it prior to deploying the new code and then re-adding it afterward, otherwise we would end up with a [failed deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#failed-deployment) due to all of the pods failing the readiness probe at the old URL. Is there any way to preserve the existing readiness URL to eliminate the need for that coordination?

pmac: I only just added the readiness URL to basket and it's not yet in prod. I'm sure we could make that happen for other projects that are already using readiness. The odd thing will be that for watchman the base URL is the full check (db, cache, etc.) and the `base/ping/` is the liveness, which I think if we use `/healthz/` as the base URL means that it would switch meanings. A small thing probably but potentially confusing for a bit. We can of course make it so that our old URLs work by simply not using their `urls.py` and adding their views directly to ours. I can't see why that wouldn't work just as well.

jgmize:
Once this ships we'll need to reconfigure the liveness and readiness probes: https://deis.com/docs/workflow/applications/managing-app-configuration/#custom-health-checks

jgmize:
Per conversation in IRC, further changes are incoming on this prior to merge. @pmac is going to update the URLs to match our current convention:  /healthz/ should be the simple check that just returns a 200, and /readiness/ should be the full check of the DB, Cache, etc.

pmac:
@jgmize and this update is now done. Thanks for taking a look!

basket/settings.py
jgmize: Now that you've [changed the URLs](https://github.com/mozmeao/basket/pull/64/commits/4ac7bd14a1cf3e7e80e8743e4f94d224fce5153f) back you need to revert this.

pmac: Ah! Great catch. Done.

glogiotatidis:
Nice find @pmac. Looks like someting we should implement in our other websites as well.

PR #77, opened 2018-05-18
Add support for `createDate` in the fxa_verified task.

### LINKED ISSUE #76: Accept a `createDate` field on the `verified` SQS message
This is to support the reconciliation between the Marketingcloud and FxA user DBs. When sending an SQS `verified` message, we want to pass in the FxA createDate so that users who have had accounts for several years do not start receiving Journeybuilder emails.

I had a quick stab at writing this myself and got lost in testing and how to parse an ISO8601 date string. I'll toss up my PR anyways, guidance or just taking over the PR would be very much appreciated.

cc @pmac

shane-tomlinson:
This is to allow reconciliation with the FxA user DB w/o
sending emails to users that have existed for several years.

cc @pmac 

fixes #76

### DIFF
--- a/basket/news/tasks.py
+++ b/basket/news/tasks.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import, unicode_literals
 
-import datetime
+from datetime import datetime, timedelta
 import logging
 from email.utils import formatdate
 from functools import wraps
@@ -31,11 +31,6 @@ from basket.news.newsletters import get_sms_vendor_id, get_transactional_message
 from basket.news.utils import (generate_token, get_accept_languages, get_best_language, get_user_data,
                                iso_format_unix_timestamp, parse_newsletters, parse_newsletters_csv,
                                SUBSCRIBE, UNSUBSCRIBE)
-
-# datetime.datetime is not a submodule, rather datetime is a class
-# within the datetime module.
-# See https://stackoverflow.com/questions/27491472/how-to-directly-import-now-from-datetime-datetime-submodule
-now = datetime.datetime.now
 
 log = logging.getLogger(__name__)
 
@@ -203,8 +198,8 @@ def et_task(func):
 
 def gmttime(basetime=None):
     if basetime is None:
-        basetime = now()
-    d = basetime + datetime.timedelta(minutes=10)
+        basetime = datetime.now()
+    d = basetime + timedelta(minutes=10)
     stamp = mktime(d.timetuple())
     return formatdate(timeval=stamp, localtime=False, usegmt=True)
 
@@ -249,7 +244,7 @@ def fxa_verified(data):
     fxa_id = data['uid']
     create_date = data.get('createDate')
     if create_date:
-        create_date = datetime.datetime.utcfromtimestamp(create_date)
+        create_date = datetime.fromtimestamp(create_date)
 
     locale = data.get('locale')
     subscribe = data.get('marketingOptIn')
--- a/basket/news/tasks.py
+++ b/basket/news/tasks.py
@@ -32,6 +32,10 @@ from basket.news.utils import (generate_token, get_accept_languages, get_best_la
                                iso_format_unix_timestamp, parse_newsletters, parse_newsletters_csv,
                                SUBSCRIBE, UNSUBSCRIBE)
 
+# datetime.datetime is not a submodule, rather datetime is a class
+# within the datetime module.
+# See https://stackoverflow.com/questions/27491472/how-to-directly-import-now-from-datetime-datetime-submodule
+now = datetime.datetime.now
 
 log = logging.getLogger(__name__)
 
@@ -197,8 +201,10 @@ def et_task(func):
     return wrapped
 
 
-def gmttime():
-    d = datetime.datetime.now() + datetime.timedelta(minutes=10)
+def gmttime(basetime=None):
+    if basetime is None:
+        basetime = now()
+    d = basetime + datetime.timedelta(minutes=10)
     stamp = mktime(d.timetuple())
     return formatdate(timeval=stamp, localtime=False, usegmt=True)
 
@@ -241,6 +247,10 @@ def fxa_verified(data):
     # used to be handled by the fxa_register view
     email = data['email']
     fxa_id = data['uid']
+    create_date = data.get('createDate')
+    if create_date:
+        create_date = datetime.datetime.utcfromtimestamp(create_date)
+
     locale = data.get('locale')
     subscribe = data.get('marketingOptIn')
     metrics = data.get('metricsContext', {})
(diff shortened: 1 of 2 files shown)


basket/news/tasks.py
```diff
@@ -197,8 +197,11 @@ def wrapped(self, *args, **kwargs):
     return wrapped
 
 
-def gmttime():
-    d = datetime.datetime.now() + datetime.timedelta(minutes=10)
+def gmttime(basetime=None):
```
shane-tomlinson: Is it reasonable to extend this function to pass in a basetime?

basket/news/tasks.py
```diff
@@ -241,6 +244,8 @@ def fxa_verified(data):
     # used to be handled by the fxa_register view
     email = data['email']
     fxa_id = data['uid']
+    create_date = data.get('createDate')
+    # TODO if create_date exists, convert it from ISO8601 to a datetime.
```
shane-tomlinson: How would you go about converting from an ISO8601 date (format `2018-05-18T13:21:54.188Z`) to a datetime? It looks like pyparsing is included, should that be used? Would you parse here or down in _update_fxa_info?

pmac: I'd use `dateutil` which is also already installed. 

```python
from dateutil.parser import parse as parse_date
...
create_date = data.get('createDate')
if create_date:
    create_date = parse_date(create_date)
```

basket/news/tasks.py
pmac: Yeah. Looks great to me. DRY :)

basket/news/tests/test_tasks.py
```diff
@@ -771,3 +786,10 @@ def test_timestamps_nothin_cached(self, cache_mock, sfmc_mock):
             'FXA_ID': data['uid'],
             'NewEmailAddress': data['email'],
         })
+
+
+class GmttimeTests(TestCase):
+    def test_basetime_provided(self):
+        basetime = parse_date('2018-05-18T13:21:54.188Z')
+        formatted_time = gmttime(basetime)
+        self.assertEqual(formatted_time, 'Fri, 18 May 2018 21:31:54 GMT')
```
shane-tomlinson: I'm not quite sure why gmttime adds 10 minutes to the time, but that's what it does.

basket/news/tests/test_tasks.py
```diff
@@ -771,3 +787,19 @@ def test_timestamps_nothin_cached(self, cache_mock, sfmc_mock):
...
+
+    def test_basetime_not_provided(self):
```
shane-tomlinson: @pmac - does this test provide any value? The fact that the test could fail makes me nervous.

shane-tomlinson:
@pmac - this is ready for review!

basket/news/tests/test_tasks.py
pmac: I think the way to go here would be to mock `datetime.datetime.now`, create a datetime object as the `return_value` of the mock, then call `gmttime()` and assert that the mock was called and that the formatted time is what is expected based on your datetime object. That's the only way I can see for making this test both worthwhile and stable.

basket/news/tests/test_tasks.py
```diff
@@ -795,11 +794,10 @@ def test_basetime_provided(self):
...
 
-    def test_basetime_not_provided(self):
-        # The hope is that calling now() and the time generated
-        # within gmttime occur in the same second, though that's not
-        # guaranteed
-        now = datetime.datetime.now()
-        formatted_now_with_basetime = gmttime(now)
-        formatted_now_without_basetime = gmttime()
-        self.assertEqual(formatted_now_without_basetime, formatted_now_with_basetime)
+    @patch('datetime.datetime.now')
```
shane-tomlinson: This error popped up when trying to patch `datetime.datetime.now`: 

> exc_info = (<type 'exceptions.TypeError'>, TypeError("can't set attributes of built-in/extension type 'datetime.datetime'",), <traceback object at 0x7f4bdfb97d88>)

I just removed the test.

shane-tomlinson: See https://circleci.com/gh/mozmeao/basket/420?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

basket/news/tasks.py
```diff
@@ -241,6 +245,10 @@ def fxa_verified(data):
     # used to be handled by the fxa_register view
     email = data['email']
     fxa_id = data['uid']
+    create_date = data.get('createDate')
+    if create_date:
+        create_date = gmtime(create_date)
```
pmac: `gmtime()` returns a `struct_time` tuple, not a datetime object. You want [datetime.datetime.utcfromtimestamp](https://docs.python.org/2/library/datetime.html#datetime.datetime.utcfromtimestamp)

basket/news/tests/test_tasks.py
```diff
@@ -771,3 +786,18 @@ def test_timestamps_nothin_cached(self, cache_mock, sfmc_mock):
             'FXA_ID': data['uid'],
             'NewEmailAddress': data['email'],
         })
+
+
+class GmttimeTests(TestCase):
+    @patch('basket.news.tasks.now')
+    def test_no_basetime_provided(self, now_mock):
+        # original time is 'Fri, 09 Sep 2016 13:33:55 GMT'
```
shane-tomlinson: I'm really confused by this test. If I use JS to create a date using the timestamp 1473428035498 (milliseconds), I get a date of `Fri, 09 Sep 2016 13:33:55 GMT`, yet the output in the test is 20:43 rather than 13:43.

Even more strange, if type gmttime's code into my console, the output is:
> 'Fri, 09 Sep 2016 12:43:55 GMT'

shane-tomlinson: @pmac - any idea why the test would return such different results to running the same code in the python module?

pmac: I think it may be because I told you wrong... I think we should be using `datetime.datetime.fromtimestamp` instead of the `utcfromtimestamp` variety. In experimenting with it that seems to be the right thing. My apologies.

basket/news/tasks.py
```diff
@@ -32,6 +32,10 @@
                                iso_format_unix_timestamp, parse_newsletters, parse_newsletters_csv,
                                SUBSCRIBE, UNSUBSCRIBE)
 
+# datetime.datetime is not a submodule, rather datetime is a class
+# within the datetime module.
+# See https://stackoverflow.com/questions/27491472/how-to-directly-import-now-from-datetime-datetime-submodule
+now = datetime.datetime.now
```
pmac: OOOO right. My apologies. This works fine though. Alternatively you can just `from datetime import datetime` and then use `datetime.now()` and `datetime.fromtimestamp()` around the file. Up to you really.

shane-tomlinson:
@pmac - updated based on your feedback to use `from datetime import datetime`, much cleaner now.

pmac:
Oh hey! So sorry I let this sit for so long! Everything looks good to me now and tests are passing! Great work. Thanks again for the PR! 💯 🎉

Issue #78, opened 2018-05-31
Add support for petition signature message types to the MoFo donation queue


pmac:
Counterpart to https://github.com/mozilla/foundation.mozilla.org/pull/1563

pmac:
Fixed via #79

PR #79, opened 2018-06-01
Accept petition signature messages on the donation queue

pmac:
These will come from MoFo's CRM platform on the same queue as donations.


pmac:
This will fix #78

PR #81, opened 2018-06-27
Add MoFo newsletter subscription option to petition signature task

pmac:
The signature task will now trigger a subscription task if the data
comes in requesting a subscription.


pmac:
@Pomax fyi

pmac:
Tests pass. Going to go ahead and merge so that we can test end-to-end on dev.

Pomax:
awesome, PR filed on the foundation site to take the signup out of the client.

PR #83, opened 2018-07-05
Fix #82: Use the official SFDC country code list

### LINKED ISSUE #82: Use official SFDC list of countries for validation
Full list can be found in the [Mozilla Foundation's repo](https://github.com/mozilla/foundation.mozilla.org/blob/87bdfe71577052ac335ca0245e0708b71bfe1154/source/js/components/petition/salesforce-country-list.js).

We should also go ahead and update the `product-details` data as well.

pmac:


basket/news/forms.py
```diff
@@ -54,8 +54,7 @@ def to_python(self, value):
 
 def country_choices():
     """Upper and Lower case country codes"""
-    regions = product_details.get_regions('en-US')
-    return regions.items() + [(code.upper(), name) for code, name in regions.iteritems()]
+    return SFDC_COUNTRIES.items() + [(code.upper(), name) for code, name in SFDC_COUNTRIES.iteritems()]
```
jpetto: Just curious - why does this list need to contain both upper and lower case country codes?

pmac: This was done a while ago, but my recollection is that it was the easiest way to have it accept either uppercase or lowercase country codes.

PR #86, opened 2018-08-10
Fix bug 1358143: Use default language if subscribe language is unsupported

### LINKED ISSUE #85: Errors when a subscription comes in with an unsupported language
See https://bugzilla.mozilla.org/show_bug.cgi?id=1358143

pmac:
Will fix #85


pmac:
Thanks Jon!

PR #107, opened 2018-10-23
Add a cool home page to basket

pmac:
Instead of a useless redirect to www.m.o steal the sweet home page from snippets-service.


pmac:
Thanks @glogiotatidis 😄

PR #109, opened 2018-10-26
Ignore slack errors during deployment

### LINKED ISSUE #108: Add error handling to slack notifications in deployment
They're in bash scripts with `-e` mode, so we just need to add `|| true` to the end of the commands mostly.

pmac:
Fix #108


metadave:
:+1:

PR #115, opened 2018-11-15
Add FxA Concerts RSVP API

### LINKED ISSUE #114: Add new API endpoint for FxA concerts
These data will go directly into an SFMC data extension:

```yaml
name: FxAccounts_Concert_RSVP
fields:
  - Email: EmailAddress
  - Firefox: Text (Y or N)
  - Campaign_ID: Text 100
  - RSVP_Time: DateTime
```

pmac:
Fix #114

### DIFF
--- a/basket/news/tasks.py
+++ b/basket/news/tasks.py
@@ -751,6 +751,16 @@ def send_recovery_message_task(email):
 
 
 @et_task
+def record_fxa_concerts_rsvp(email, is_firefox, campaign_id):
+    sfmc.add_row('FxAccounts_Concert_RSVP', {
+        'Email': email,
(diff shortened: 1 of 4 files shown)


pmac:
@jpetto I've not added tests yet, but this is the idea. The call from bedrock will be something like:

```python
data = {
    'email': email_address,
    'is_firefox': 'Y', # or 'N'. for the view to figure out.
    'campaign_id': 'livenation-2018', # or something. just here in case there's another of these some day.
    'api_key': settings.BASKET_API_KEY,
}
basket.request('post', 'fxa-concerts-rsvp', data=data)
```

pmac:
Now with 100% more tests! 🎆

PR #143, opened 2019-03-29
Add API endpoint for Common Voice Goals Journey

### LINKED ISSUE #135: Store Commonvoice Contact Data
Take in Commonvoice data (either via HTTP API or Queue) and store it in SFMC. We'll discuss the details of the data to be exchanged and the method here.

/cc @gregoor

pmac:
Adds the `/news/common-voice-goals/` API endpoint.

Fix #135

### DIFF
--- a/basket/news/backends/sfdc.py
+++ b/basket/news/backends/sfdc.py
@@ -63,6 +63,10 @@ FIELD_MAP = {
     'fsa_city': 'MailingCity',
     'fsa_current_status': 'FSA_Current_Status__c',
     'fsa_allow_share': 'FSA_Allow_Info_Shared__c',
+    'cv_days_interval': 'cv_days_interval__c',
+    'cv_created_at': 'cv_created_at__c',
+    'cv_goal_reached_at': 'cv_goal_reached_at__c',
+    'cv_first_contribution_date': 'cv_first_contr_dt__c',
 }
 INV_FIELD_MAP = {v: k for k, v in FIELD_MAP.items()}
 FIELD_DEFAULTS = {
--- a/basket/news/forms.py
+++ b/basket/news/forms.py
@@ -57,12 +57,15 @@ def country_choices():
     return SFDC_COUNTRIES.items() + [(code.upper(), name) for code, name in SFDC_COUNTRIES.iteritems()]
 
 
+UTC_DATETIME_FORMATS = ['%Y-%m-%dT%H:%M:%SZ']
+
+
 class CommonVoiceForm(forms.Form):
     email = EmailField()
     days_interval = forms.IntegerField(required=False)
-    created_at = forms.DateTimeField(required=False)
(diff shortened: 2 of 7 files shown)


pmac:
This isn't finished yet, but can be picked up by someone else if this work needs to be done sooner than I'm able. It still needs some tests and some verification with the CV folks for datetime formats they'll be sending. Depending on that we might have to adjust the formats accepted by the `DateTimeField`s. 

For testing this when it lands on dev a curl command like the following should work:

```bash
$ curl --data-urlencode "email=cv-test-20190329@restmail.net" -d "days_interval=7" -d "created_at=2019-03-29 14:54:47" -d "first_contribution_date=2019-03-19 14:54:47" -H "x-api-key: <valid-basket-dev-api-key>" https://basket-dev.allizom.org/news/common-voice-goals/
```

pmac:
Okay. We need to get this going again yes? What's the deadline and when can we start testing. 

@Gregoor is it possible for you to paste here example data in the formats you expect to use? I'm mostly curious about the exact datetime format to expect.

Gregoor:
Sorry, just got back from vacation. The datetime format I'd expect would be `2019-04-15T06:03:54Z`, but really I'm flexible and would just use whatever's easiest for you to implement.

Gregoor:
Okay here's some examples:

Re-Engagement emails:
`cv_first_contribution_date: "2019-04-15T06:03:54Z"`

Goals:
`cv_created_at: "2019-04-07T12:42:34Z"`
`cv_days_interval: 7`
`cv_goal_reached_at: null`

In that case an email should've been sent at `2019-04-07T12:42:34Z` + 3.5 days

pmac:
Okay. I've added some updates and all is looking well. We should get this onto dev for some testing.

pmac:
I just went ahead with the merge as this is all new stuff and needs to be tested (in the morning) on dev. 

@Gregoor I updated the date format to match what you gave me, which conveniently is exactly what Salesforce wants, so I really just validate and pass it through. The following request should work:

```bash
$ curl --data-urlencode "email=cv-test-20190415@restmail.net" -d "days_interval=7" -d "created_at=2019-04-15T14:54:47Z" -d "first_contribution_date=2019-04-14T14:54:47Z" -H "x-api-key: <valid-basket-dev-api-key>" https://basket-dev.allizom.org/news/common-voice-goals/
```

pmac:
And then you can look up the added or updated user's data using the same API key with this command:

```bash
$ curl  -H "x-api-key: <valid-basket-dev-api-key>" "https://basket-dev.allizom.org/news/lookup-user/?email=cv-test-20190415@restmail.net"
```

That will return JSON data for the user that is directly returned from Salesforce. You'll see the fields prefixed with `cv_` in the data.

Gregoor:
I've sent a couple of request over that came back with 200 :)

PR #153, opened 2019-04-12
Add subscribe event type for donations queue

### LINKED ISSUE #152: adding a new event type to basket
For the foundation website's newsletter(s) signup functionality, we'd like to send signup events on a new event type. We're currently using `crm_petition_data` for mofo petitions, but we'd like to also have a `newsletter_signup_data` event type for dealing with (comma-separated) newsletter(s) that users can register for.

pmac:
Fix #152

### DIFF
--- a/basket/news/management/commands/process_donations_queue.py
+++ b/basket/news/management/commands/process_donations_queue.py
@@ -15,6 +15,7 @@ from raven.contrib.django.raven_compat.models import client as sentry_client
 from basket.news.tasks import (
     process_donation,
     process_donation_event,
+    process_newsletter_subscribe,
     process_petition_signature,
 )
 
@@ -22,6 +23,7 @@ from basket.news.tasks import (
 EVENT_TYPES = {
     'donation': process_donation,
     'crm_petition_data': process_petition_signature,
+    'newsletter_signup_data': process_newsletter_subscribe,
     'DEFAULT': process_donation_event,
 }
 
--- a/basket/news/tasks.py
+++ b/basket/news/tasks.py
@@ -31,7 +31,7 @@ from basket.news.models import (FailedTask, Newsletter, Interest,
 from basket.news.newsletters import get_sms_vendor_id, get_transactional_message_ids, newsletter_map
 from basket.news.utils import (generate_token, get_accept_languages, get_best_language, get_user_data,
                                iso_format_unix_timestamp, parse_newsletters, parse_newsletters_csv,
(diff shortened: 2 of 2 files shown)


pmac:
@Pomax FYI

basket/news/tasks.py
```diff
@@ -928,6 +928,12 @@ def process_donation(data):
             raise
 
 
+@et_task
+def process_newsletter_subscribe(data):
+    data['lang'] = get_best_supported_lang(data['lang'])
```
jpetto: Is there any need to ensure `lang` exists in `data`? Or that `data` contains all info needed to `upsert_user`?

basket/news/management/commands/process_donations_queue.py
```diff
@@ -15,13 +15,15 @@
...
     process_donation_event,
+    process_newsletter_subscribe,
     process_petition_signature,
 )
 
 
 EVENT_TYPES = {
     'donation': process_donation,
     'crm_petition_data': process_petition_signature,
+    'newsletter_signup_data': process_newsletter_subscribe,
```
jpetto: Out of curiosity, where are these used?

basket/news/tasks.py
pmac: It should all always be there. See the code on the mofo side:

https://github.com/mozilla/foundation.mozilla.org/blob/060922af4dd07f35f8119f2596228f47946b4487/network-api/networkapi/campaign/views.py#L87-L93

I really only did the `best_supported_lang` call to prepare for them someday sending a real language. It's hard coded to `en` for now.

basket/news/management/commands/process_donations_queue.py
pmac: What are the `these` in this question? These `EVENT_TYPES`? If so they're used just a few lines below this in this file.

pmac:
The details can be found in the linked issue (#152) where the code on the foundation's site can be found. I linked it above.

basket/news/management/commands/process_donations_queue.py
jpetto: Ah, I see. I did a search for the types and only found `donation`. The other types are specified in the incoming data. Got it.

jpetto:
All good. Apologies for not reading the related issue fully. :shipit: !!

pmac:
@jpetto oh no need for apology. I should have pointed all of that out in the PR description. Thanks for being thorough!

PR #165, opened 2019-05-08
Add two new fields for Common Voice endpoint

### LINKED ISSUE #164: Additional Common Voice fields (2 day streak & Last Active date)
There are 2 new fields that we need ported over from Common Voice into SFDC. The fields are: 

2 day streak / boolean (cv_two_day_streak__c)
Common Voice Last Active Date (date) (cv_last_active_dt__c)

Please work with Desiree and Gregor w/ the Common Voice team to get these data values tested in the ADVO SFDC Sandbox, then moved to Production.

pmac:
Fix #164

### DIFF
--- a/basket/news/forms.py
+++ b/basket/news/forms.py
@@ -78,6 +78,10 @@ class CommonVoiceForm(forms.Form):
     first_contribution_date = forms.CharField(required=False,
                                               max_length=20,
                                               validators=[validate_datetime_str])
+    last_active_date = forms.CharField(required=False,
+                                       max_length=20,
+                                       validators=[validate_datetime_str])
+    two_day_streak = forms.NullBooleanField()
 
 
 class SubscribeForm(forms.Form):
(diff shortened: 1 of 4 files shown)


pmac:
@Gregoor these are the new fields. Sending booleans to the `two_day_streak` field should be the string `True` or `False`. The `last_active_date` field is the same date format as the others.

Gregoor:
Excellent, sounds good 👌 Thanks!

basket/news/forms.py
```diff
@@ -78,6 +78,10 @@ class CommonVoiceForm(forms.Form):
     first_contribution_date = forms.CharField(required=False,
                                               max_length=20,
                                               validators=[validate_datetime_str])
+    last_active_date = forms.CharField(required=False,
+                                       max_length=20,
+                                       validators=[validate_datetime_str])
+    two_day_streak = forms.NullBooleanField()
```
jpetto: Do you think it's worth testing this value as `Null`/an empty string?

pmac: The test that doesn't include that value is kindof a test for that. It's empty, which means that field will be `None` which is then filtered out and not sent to the task at all. I think that should cover pretty well. Could possibly be better, but this view is protected by API key so we can better expect good values from our partner on this one as well.

pmac:
Thanks Jon!

Issue #161, opened 2019-05-03
Include FxA Tips subscription info with user data


pmac:
Include the FxA tips subscirption status of an email address with requests to `lookup-user`. This data is in data extensions in SFMC. The user must be in the FxA account holders extension, and not in the FxA account opt-out extension. The data returned can have 3 states for this. 

1. Nothing returned for `fxa-tips` means that the user has no record of having an FxA.
2. `True` means the user is an FxA holder, but has no opt-out.
3. `False` means the user is an FxA holder, but has an opt-out.

ldwright:
Please coordinate with Lauren Niolet on the plan surrounding FxA Tips.

_Based on the current plan to my knowledge, we are sunsetting the existing Firefox Accounts Tips subscription, and there will not be any sort of new FxA Tips subscription stood up. See doc below._

https://docs.google.com/spreadsheets/d/1CdsuwXBhXhJ9LBMthRMDx9ATEGWDiKXlZoPFWLjjKzI/edit#gid=0

pmac:
Right. I was trying to refer to the FxA account holders emails that currently come only from SFMC. That's what this issue is about. The current newsletter should be unaffected by this work.

pmac:
Updates: We will not need to indicate whether the user has opted out of these emails because the opt-out mechanism is being removed and opt-out will now only be possible via FxA deletion. So this will only need to return something like `'fxa-user': True` or `False`.

pmac:
We should also make inclusion of that data optional via a query param to these requests. We'll sometimes be able to know if they're an FxA user via other means, so there will be no need to take the time to look them up in SFMC in those cases.

Issue #160, opened 2019-04-29
Record source_url for all FxA login events


pmac:
Currently only recording it for users that opt into the newsletter.

PR #169, opened 2019-05-15
Add return of FxA holder status with user data requests

pmac:
Adding a `?fxa=1` to requests to /news/user/ or /news/lookup-user/ will now return a `has_fxa` parameter. Also record FxA source URL regardless of newsletter optin.

Re #160 #161

### DIFF
--- a/basket/news/views.py
+++ b/basket/news/views.py
@@ -550,7 +550,8 @@ def user(request, token):
             data['email'] = email
         return update_user_task(request, SET, data)
 
-    return get_user(token)
+    get_fxa = 'fxa' in request.GET
+    return get_user(token, get_fxa=get_fxa)
 
 
 @require_POST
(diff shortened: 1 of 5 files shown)


basket/news/views.py
```diff
@@ -685,6 +686,7 @@ def lookup_user(request):
 
     token = request.GET.get('token', None)
     email = request.GET.get('email', None)
+    get_fxa = 'fxa' in request.GET
```
jpetto: Just to clarify, we don't care what the value of the `fxa` key is in the URL, only that the key is present, right?

pmac: I think that's the case. Bedrock will be sending `?fxa=1` but really it's just acting as a boolean indicating a desire to have the fxa status returned. I don't think we need to support an off state like `?fxa=0` or anything like that.

PR #171, opened 2019-05-15
Add indent option to newsletter admin and data

### LINKED ISSUE #170: Add indention setting to newsletters in admin
Some newsletters need to appear indented in the email preference center on bedrock. We will add a boolean field to the newsletter model to allow that indentation to be controlled via the bedrock admin.

pmac:
Fix #170


pmac:
This change is pretty trivial. Going to go ahead and merge so LCM can start trying it out on my bedrock demo and basket stage.

PR #173, opened 2019-05-21
Accept `newsletters` data with FxA Verified events

### LINKED ISSUE #172: Update FxA integration to support multiple newsletter subscriptions
We currently accept only a boolean field in the `verified` queue messages: `marketingOptIn`. That mapped to a single pre-defined newsletter. They'd like to be able to send the list of newsletters being opted into in the message. We'll determine the changes in this issue and proceed from there.

Re https://github.com/mozilla/fxa/issues/1102

/cc @vbudhram

pmac:
Fix #172

### DIFF
--- a/basket/news/tasks.py
+++ b/basket/news/tasks.py
@@ -248,6 +248,7 @@ def fxa_verified(data):
 
     locale = data.get('locale')
     subscribe = data.get('marketingOptIn')
+    newsletters = data.get('newsletters')
     metrics = data.get('metricsContext', {})
     service = data.get('service', '')
     country = data.get('countryCode', '')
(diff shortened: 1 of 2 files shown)


basket/news/tasks.py
```diff
@@ -248,6 +248,7 @@ def fxa_verified(data):
 
     locale = data.get('locale')
     subscribe = data.get('marketingOptIn')
+    newsletters = data.get('newsletters')
```
jpetto: Just making sure `newsletters` doesn't need a default value. If not, this looks good to go to me!

pmac: the default value returned from `get()` is `None`, and below I'm just doing a check to see if it's truthy, which would fail if it were none or an empty list (which is what FxA will send if people select no newsletters). So I believe we're good, and there are tests of this method without the `newsletters` key.

pmac:
Thanks Jon!

PR #177, opened 2019-05-22
Switch to official Salesforce FuelSDK libraries

### LINKED ISSUE #98: Upgrade FuelSDK
https://pypi.org/project/Salesforce-FuelSDK/

This also includes an upgrade to the suds library.

pmac:
Fix #98

### DIFF
--- a/basket/news/tests/test_middleware.py
+++ b/basket/news/tests/test_middleware.py
@@ -12,7 +12,9 @@ def test_is_ip_address():
     assert not is_ip_address('42.basket.mozilla.org')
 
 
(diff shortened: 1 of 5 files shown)


jgmize:
needs rebase but other than that looking good so far; pulling down locally for testing

pmac:
Yeah. I think it needs some more investigation and testing, but so far so good. Mostly wanted circle to run, but Github is having issues with that at present.

pmac:
We were using my fork of the Fuel-SDK library because of some shortcomings of the official library. It looks like we're going to have to continue doing that. I was installing from my personal fork, but I'll fork it in `mozmeao` and install from there.

pmac:
Okay. I think this is ready for a full test and review.

PR #181, opened 2019-05-23
Upgrade Python to 3.7

### LINKED ISSUE #179: Upgrade Python to 3.7
Prerequisite for #97

pmac:
Fix #179


pmac:
Tests are passing. Probably needs more thorough manual testing, but should be ready for a code review.

basket/news/country_codes.py
```diff
@@ -252,7 +252,7 @@
   'zm': 'Zambia',
   'zw': 'Zimbabwe',
 }
-SFDC_COUNTRIES_LIST = SFDC_COUNTRIES.keys()
+SFDC_COUNTRIES_LIST = list(SFDC_COUNTRIES.keys())
```
jpetto: Does this need to be converted to a list because it's exported/imported?

pmac: Dict functions like `keys()` and `items()` return an iterator in Python 3 where they used to return a `list` in Python 2. The 2to3 tool did this just to make sure it all worked the way it used to. I removed a lot of those uses of `list()` for this PR but left that one since I wasn't immediately sure how it was used. It would probably be fine to not cast it here, but it also doesn't hurt.

PR #184, opened 2019-05-23
Set MySQL Strict Mode

### LINKED ISSUE #183: Use MySQL Strict Mode
I noticed the following warning during DB migrations in a Jenkins deployment:

> WARNINGS:
> 
> ?: (mysql.W002) MySQL Strict Mode is not set for database connection 'default'
> 
> 	HINT: MySQL's Strict Mode fixes many data integrity problems in MySQL, such as data truncation upon insertion, by escalating warnings into errors. It is strongly recommended you activate it. See: https://docs.djangoproject.com/en/1.11/ref/databases/#mysql-sql-mode

Seems like a good idea to do that.

pmac:
Fix #183

### DIFF
--- a/basket/settings.py
+++ b/basket/settings.py
@@ -55,7 +55,10 @@ DATABASES = {
                       cast=dj_database_url.parse),
 }
 if DATABASES['default']['ENGINE'] == 'django.db.backends.mysql':
-    DATABASES['default']['OPTIONS'] = {'init_command': 'SET storage_engine=InnoDB'}
+    DATABASES['default']['OPTIONS'] = {
+        'init_command': "SET sql_mode='STRICT_TRANS_TABLES'",
+    }
+
 
 CACHES = {
     'default': config('CACHE_URL',


pmac:
Thoughts @mozmeao/theconn ?

jpetto:
Seems as though we should definitely specify one of the strict modes.

Do we have any concern about using `STRICT_ALL_TABLES` in regards to the effects on [nontransactional tables](https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sql-mode-important)? (I'm not sure which kind we're using in basket.)

> For STRICT_ALL_TABLES, MySQL returns an error and ignores the rest of the rows. However, because the earlier rows have been inserted or updated, the result is a partial update. To avoid this, use single-row statements, which can be aborted without changing the table.

pmac:
Updated. We're using a new enough Mariadb that the default storage engine is already InnoDB, so we just need to set this to be up to date. And we should probably look at upgrading maria or finally moving the Postgres.

PR #185, opened 2019-05-24
Integrate with FxA Oauth to get email and redirect to Email Prefs

### LINKED ISSUE #167: Add new endpoint for FxA to lookup users and redirect to preference center
We're still working out whether this will need to accept only an FxA ID, or also another random token, but it will need to do the following:

1. Look up the user using the token provided in SFMC FxA table
2. Using the email returned, look up the user in SFDC
    * If no SFDC record found for user, create one, and redirect to the email pref center on bedrock using the new token
    * If SFDC record is found, use the token in that record to redirect to the email pref center on bedrock
3. Include a query param (e.g. `?fxa=1`) on the links to the email pref center so that a new lookup of the user in SFDC is not required. The request for user data in that case should not ask for FxA status.

pmac:
This is preliminary. I need to add tests and things, but this works locally and I wanted some feed back.

Fix #167 

/cc @rfk

### DIFF
--- a/basket/news/views.py
+++ b/basket/news/views.py
@@ -1,8 +1,10 @@
+import os
 import re
 from time import time
 
 from django.conf import settings
 from django.core.exceptions import NON_FIELD_ERRORS
+from django.http import HttpResponseRedirect
 from django.shortcuts import render
 from django.views.decorators.cache import cache_page, never_cache
 from django.views.decorators.csrf import csrf_exempt
@@ -12,6 +14,7 @@ from basket import errors
 from django_statsd.clients import statsd
 from ratelimit.exceptions import Ratelimited
 from ratelimit.utils import is_ratelimited
+from raven.contrib.django.raven_compat.models import client as sentry_client
 from synctool.routing import Route
 
 from basket.news.forms import CommonVoiceForm, SubscribeForm, UpdateUserMeta, SOURCE_URL_RE
@@ -41,6 +44,7 @@ from basket.news.utils import (
     get_accept_languages,
     get_best_language,
     get_best_supported_lang,
+    get_fxa_clients,
     get_user_data,
     get_user,
     has_valid_api_key,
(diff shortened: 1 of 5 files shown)


basket/news/views.py
```diff
@@ -119,6 +122,66 @@ def ratelimited(request, e):
...
+    state = request.GET.get('state')
+    if not (code and state):
+        # need a better error state here
+        return HttpResponseJSON({
+            'status': 'error',
+            'message': 'no code or state returned',
+        }, 400)
+
+    if fxa_state != state:
+        # need a better error state here
```
rfk: I recommend clearing `request.session['fxa_state']` here if possible, otherwise an attacker gets unlimited guesses at the correct state token (and might even be able to use timing information to reduce the complexity of guessing). It's *probably* fine if you can't conveniently clear it,  because the attack is "I can force you to login to my email preferences page" and that doesn't seem like a worthwhile thing for an attacker to achieve.

pmac: Ah yes! Good catch. Meant to do that.

basket/news/views.py
```diff
@@ -119,6 +122,69 @@ def ratelimited(request, e):
...
+                                     email=request.GET.get('email'))
+    return HttpResponseRedirect(url)
+
+
+@require_safe
+def fxa_success(request):
+    fxa_state = request.session.get('fxa_state')
+    if fxa_state:
+        # state is one-time use
+        del request.session['fxa_state']
```
pmac: @rfk this should be fine right? I can't think of a reason to need to request it again.

pmac:
@jpetto please take a look at this as soon as you can. Need to get it up on staging for testing asap. I still need to figure out the best way to handle errors and where to send people that encounter them. We may need a new page specifically for this...

basket/news/views.py
```diff
@@ -119,6 +123,85 @@ def ratelimited(request, e):
...
+        user_profile = fxa_profile.get_profile(access_token)
+    except Exception:
+        sentry_client.captureException()
+        # need a better error state here
+        return HttpResponseJSON({
+            'status': 'error',
+            'message': 'error communicating with FxA',
+        }, 500)
+
+    # TODO use user_profile['uid'] to check account in SFMC
```
jpetto: Just verifying - this TODO is to check for an existing record/token in SFMC, and if found, use that token for the `get_user_data` call below?

jpetto:
@pmac One question on a TODO, and we should have tests for the new views, but I don't see anything stopping us from getting this up for actual end to end testing.

basket/news/views.py
pmac: This is a note for a nice-to-have thing. In SFMC we have a table with email and FxA_ID. This is updated via messages we receive from FxA on a queue when people verify their email on a new account, but it's not the source of record and no sync process is perfect. So this is an opportunity to verify and correct our record of these users. I don't think we have to do that, but we could and maybe should. It's just not important for v1 of this and time is short.

pmac:
Thanks Jon! I'm going to work on some tests for this, but in a separate PR. I'll merge this now so that it can get out to stage for some real user testing.

PR #191, opened 2019-06-04
Docker compose and multi stage

pmac:
Updates for docker and docker-compose and using those in CI. Adds multi-stage docker build resulting in a smaller and more secure final image to deploy, and uses docker-compose for all building and test running in CI.


pmac:
@jpetto now that Py3k has landed I believe this is ready for r?

jpetto:
Note: let's get those docs updated soon. 😉

PR #192, opened 2019-06-05
[fix #166] REST endpoint for Subhub to populate Salesforce

### LINKED ISSUE #166: New REST endpoint for Subscription processor to CRM
We will be posting webhook events from payment provider to subscription processor to CRM. Here is a sample request:

```json
{
    "type" : "charge.succeeded",
    "payload" : { "id" : "someid", "someotherfield" : "someothervalue"}
}
```

The payload is generic, the Basket program will have to parse based on the type. The definition for each type is currently being worked on. We are assuming this should affect the signature of the REST endpoint we need from Basket.

jpetto:
Create a simple API to capture Stripe transactions and put them in SFDC/SFMC.

All data will come to a single view/endpoint. That view will check the API token, then route the payload to the appropriate task based on the `event_type` sent in the payload. Tasks perform various bits of data transformation and logic before inserting/updating records in Salesforce.

Fixes #166

### DIFF
--- a/basket/news/tasks.py
+++ b/basket/news/tasks.py
@@ -718,6 +718,157 @@ def record_source_url(email, source_url, newsletter_id):
 
 
 @et_task
+def process_subhub_event_customer_created(data):
+    """
+    Creates or updates a SFDC customer when a new Stripe customer is created
+    """
+    statsd.incr('news.tasks.process_subhub_event.customer_created')
+
+    first, last = split_name(data['name'])
+    contact_data = {
+        'fxa_id': data['user_id']
+        'stripe_id': data['customer_id']
+    }
+    user_data = get_user_data(email=data['email'])
+
+    # if user was found in sfdc, see if we should update their name(s)
+    if user_data:
+        # if current last name is '_', update it
+        if user_data['last_name'] == '_':
+            contact_data['last_name'] = last
+
+        # if current last name is blank/Null, update it
+        if not user_data['first_name']:
+            contact_data['first_name'] = first
+
+        if 'first_name' in contact_data or 'last_name' in contact_data:
+            sfdc.update(user_data, contact_data)
+    # if no user was found, create new user in sfdc
+    else:
+        contact_data['first_name'] = first
+        contact_data['last_name'] = last
+        contact_data['email'] = data['email']
+
+        # create the user in sfdc
+        sfdc.add(contact_data)
+
+
+@et_task
+def process_subhub_event_customer_updated(data):
+    """
+    Updates customer record in SFDC if Stripe customer changes their name or
+    email address
+    """
+    statsd.incr('news.tasks.process_subhub_event.customer_updated')
+
+    first, last = split_name(data['name'])
+    contact_data = {}
+
+    # find user by their Stripe ID
+    user_data = get_user_data(stripe_id=data['customer_id'])
+
+    # if user was found in sfdc, see if we should update their name(s)
+    if user_data:
+        # if current last name is '_', update it
+        if user_data['last_name'] == '_':
+            contact_data['last_name'] = last
+
+        # if current last name is blank/Null, update it
+        if not user_data['first_name']:
+            contact_data['first_name'] = first
+
+        if 'first_name' in contact_data or 'last_name' in contact_data:
+            sfdc.update(user_data, contact_data)
+
+    # TODO: do we need to do anything if the customer wasn't found?
+
+
+@et_task
+def process_subhub_event_subscription_charge(data):
+    """
+    This method handles both new and recurring charges. New charges will not
+    have data for the current_period_start and current_period_end fields.
+
+    There are two events that trigger this task:
+    - invoice.finalized
+    - invoice.payment_succeeded
+
+    Each of these events contains different information on the charge which
+    will need to be combined. So we need to check for an existing record
+    prior to doing any inserts or updates.
+
+    invoice.finalized *should* be sent first, which is missing the following
+    credit card related fields:
+
+    - brand
+    - last4
+    - exp_month
+    - exp_year
+    """
+
+    # determine if new or recurring payment
+    recurring_charge = True if data['current_period_start'] and data['current_period_end'] else False
+
+    if recurring_charge:
+        statsd.incr('news.tasks.process_subhub_event.new_subscription_charge')
+    else:
+        statsd.incr('news.tasks.process_subhub_event.recurring_subscription_charge')
+
+    # determine which event we're receiving
+    event = 'succeeded' if 'brand' in data and 'last4' in data else 'finalized'
+
+    # try to find the opportunity (in case one of the events already fired)
+    user_data = get_user_data(stripe_id=data['customer_id'])
+
+    if user_data:
+        # see if an opportunity exists with the provided charge_id
+        charge_id = data['charge_id']
+
+        transaction_data = {
+            'Stripe_Cust_Id__c': data['customer_id'],
+            'PMT_Subscription_ID__c': data['subscription_id'],
+            'CloseDate': iso_format_unix_timestamp(data['created']),
+            'Billing_Cycle_End__c': iso_format_unix_timestamp(data['period_end']),
+            'Billing_Cycle_Start__c': iso_format_unix_timestamp(data['period_start']),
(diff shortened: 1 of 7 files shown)


jpetto:
@pmac Still very much in draft mode, but would love a sanity check when you get a chance. 😄

basket/news/views.py
```diff
@@ -939,3 +955,36 @@ def update_user_task(request, api_call_type, data=None, optin=False, sync=False)
...
+    else:
+        etype = data['event_type']
+        processor = SUBHUB_EVENT_TYPES.get(etype)
+
+        if processor:
+            processor.delay(data)
+            return HttpResponseJSON({'status': 'ok'})
+        else:
+            # TODO: set a better HTTP response code? 409?
+            return HttpResponseJSON({'status': 'unknown event type'})
```
pmac: At least return a 400 and a status of `error` and this as `desc`. This is good though 👍

basket/news/views.py
```diff
@@ -939,3 +955,36 @@ def update_user_task(request, api_call_type, data=None, optin=False, sync=False)
...
+    if not has_valid_api_key(request):
+        return HttpResponseJSON({
+            'status': 'error',
+            'desc': 'requires a valid API-key',
+            'code': errors.BASKET_AUTH_ERROR,
+        }, 401)
+
+    try:
+        data = json.loads(request.body)
+    except ValueError as _:
```
pmac: If you don't need the exception object just leave the `as _` off.
```suggestion
    except ValueError:
```

basket/news/views.py
```diff
@@ -939,3 +955,36 @@ def update_user_task(request, api_call_type, data=None, optin=False, sync=False)
...
+        statsd.incr('subhub_post.message.json_error')
+        sentry_client.captureException(data={'extra': {'request.body': request.body}})
```
pmac: 👍

basket/news/utils.py
```diff
@@ -280,9 +280,10 @@ def language_code_is_valid(code):
 ]
 
 
-def get_user_data(token=None, email=None, extra_fields=None, get_fxa=False):
+def get_user_data(token=None, email=None, stripe_id=None, extra_fields=None, get_fxa=False):
```
pmac: Still working on making this work I guess?

basket/news/tasks.py
```diff
@@ -717,6 +717,157 @@ def record_source_url(email, source_url, newsletter_id):
...
+            transaction_data['Credit_Card_Exp__c'] = data['exp_year']  # TODO: check w/marty on SF field names here - they are duplicated in the doc
+
+        try:
+            # will raise a SalesforceMalformedRequest if not found
+            sfdc.opportunity.update('PMT_Transaction_ID__c/{}'.format(charge_id), transaction_data)
+            # TODO: is this the proper syntax? unsure about the use/purpose of the first param here - is there some magic happening that makes this a key?
+        except sfapi.SalesforceMalformedRequest as e:
+            sfdc.opportunity.create(transaction_data)
+    else:
+        # TODO: do we need to do anything here if the user wasn't found?
```
pmac: Maybe there's enough info in the data to create the user if none is found? Really only need email address.

basket/news/utils.py
jpetto: Yeah - just put this in as a starting point.

basket/news/tasks.py
jpetto: According to the doc, the only user identifiable info in these payloads is the Stripe ID, so I don't think we can do much?

basket/news/tasks.py
```diff
@@ -717,6 +717,157 @@ def record_source_url(email, source_url, newsletter_id):
...
+        if event == 'succeeded':
+            transaction_data['Credit_Card_Type__c'] = data['brand']
+            transaction_data['Last_4_Digits__c'] = data['last4']
+            transaction_data['Credit_Card_Exp__c'] = data['exp_month']
+            transaction_data['Credit_Card_Exp__c'] = data['exp_year']  # TODO: check w/marty on SF field names here - they are duplicated in the doc
+
+        try:
+            # will raise a SalesforceMalformedRequest if not found
+            sfdc.opportunity.update('PMT_Transaction_ID__c/{}'.format(charge_id), transaction_data)
+            # TODO: is this the proper syntax? unsure about the use/purpose of the first param here - is there some magic happening that makes this a key?
```
jpetto: Would love some input on this bit. 😄

pmac: It's the lookup basically. The first bit is the filed to look in, and the second is the data.

https://github.com/simple-salesforce/simple-salesforce#other-options

jpetto:
@glogiotatidis Do you have time to review this during your morning tomorrow? Tests are only stubs at the moment, but it would be great to get this up on dev for end to end testing of the Salesforce API.

basket/news/utils.py
```diff
@@ -280,9 +280,10 @@ def language_code_is_valid(code):
 ]
 
 
-def get_user_data(token=None, email=None, extra_fields=None, get_fxa=False):
+def get_user_data(token=None, email=None, payee_id=None, extra_fields=None, get_fxa=False):
```
glogiotatidis: This should be fine but generally speaking I'm careful adding more parameters out of order, especially in a function with all optional parameters. I did a quick search and the only place where we don't explicitly name the parameter is in https://github.com/mozmeao/basket/blob/79624a385bcef27c639c312916282b02d11edac6/basket/news/tasks.py#L386, so maybe you can change that to `get_user_data(token=data.get('token'), email=data.get('email'), extra_fields=['id']))` just to be on the safe side.

basket/news/tasks.py
```diff
@@ -717,6 +717,250 @@ def record_source_url(email, source_url, newsletter_id):
...
+    # processor/Stripe ID, so if the user wasn't found by that, there's really
+    # nothing to be done here
+    if user_data:
+        # this is the primary key for the opportunity
+        charge_id = data['charge_id']
+
+        # just in case amount data is missing/malformed
+        try:
+            # amount values are cents - convert to dollars
+            amount_paid = int(data['amount']) / 100
```
glogiotatidis: This always returns `int`. 
```python
int(108.5) / 100 = 1
```
Just a heads up, not sure if that's what you want here.

basket/news/tasks.py
```diff
@@ -717,6 +717,250 @@ def record_source_url(email, source_url, newsletter_id):
...
+            # set properly in the except block below
+            sfdc.opportunity.update('PMT_Transaction_ID__c/{}'.format(charge_id), transaction_data)
+        except sfapi.SalesforceMalformedRequest:
+            # we are creating a new opportunity, so put the charge_id in
+            transaction_data['PMT_Transaction_ID__c'] = charge_id
+
+            # determine if this is an initial or recurring payment
+            try:
+                # look for existing opportunities with the current subscription_id
+                _ = sfdc.opportunity.get_by_custom_id('PMT_Subscription_ID__C', data['subscription_id'])
```
glogiotatidis: Why do you need `_`?

basket/news/utils.py
jpetto: Ah, nice catch!

I did think maybe `get_user_data` needs a refactor - maybe taking the (now) three identifiers and expecting a dict instead - something like `{ 'type': 'token', 'value': '123456' }`. Anyway, I'll leave that for later clean up.

basket/news/tasks.py
jpetto: Meant to get rid of that - was just coding quickly and wasn't sure if I was going to be using the result. Will remove.

basket/news/tasks.py
jpetto: Hmmm, good catch. Will replace with:

`int(data['amount']) / float(100)`

That should do it, right?

jpetto:
@glogiotatidis Made a commit addressing your feedback. If you have time to verify tomorrow morning, that would be great. Hopefully we can push this out to dev tomorrow. 😁

basket/news/tasks.py
glogiotatidis: Yes that works. Another trick is

`int(data['amount']) / 100.0`

glogiotatidis:
Looks good, thanks!

PR #194, opened 2019-06-13
Fix #193 - handle NoneType in last_name data

### LINKED ISSUE #193: last_name data resolving as NoneType
We recently started getting data from MoFo that's throwing errors when attempting to pull `last_name` from the data.

[This line](https://github.com/mozmeao/basket/blob/prod/basket/news/backends/sfdc.py#L332): `if not data.get('last_name', '').strip():`

Is resulting in this error: `AttributeError: 'NoneType' object has no attribute 'strip'`

Related: MoFo [pushed a code change on 6/6 defaulting names to `None`](https://github.com/mozilla/foundation.mozilla.org/commit/10d2a12881a237cee7baaaf9cbb319715e7b8363#diff-d66144df8fec88b46d1f5e779b75122fR93). This is likely what revealed the bug.

jpetto:
Fixes #193

### DIFF
--- a/basket/news/backends/sfdc.py
+++ b/basket/news/backends/sfdc.py
@@ -329,7 +329,7 @@ class SFDC(object):
         @param data: user data to add as a new contact.
         @return: None
         """
-        if not data.get('last_name', '').strip():
+        if not (data.get('last_name', '') or '').strip():
             data['last_name'] = LAST_NAME_DEFAULT_VALUE
         self.contact.create(to_vendor(data))
 


basket/news/backends/sfdc.py
```diff
@@ -329,7 +329,12 @@ def add(self, data):
         @param data: user data to add as a new contact.
         @return: None
         """
-        if not data.get('last_name', '').strip():
+        try:
+            last_name = data.get('last_name', '').strip()
+        except AttributeError:
+            last_name = None
```
jgmize: This works, but here's another, simpler approach:

```python
last_name = (data.get('last_name', '') or '').strip()
```

PR #196, opened 2019-06-14
Handle TypeError when field value is None. Fix #195

### LINKED ISSUE #195: None value throwing errors in sfdc.py/to_vendor
Appears to be similar to #193. Changes to MoFo are impacting collecting subscriptions.

Specific error is [this line in sfdc.py in the `to_vendor` method](https://github.com/mozmeao/basket/blob/prod/basket/news/backends/sfdc.py#L165):

`if field in contact and len(contact[field]) > length:`

Error thrown is: `TypeError: object of type 'NoneType' has no len()`

Sentry shows `FirstName` has a value of `None`.

Error report is [here](https://sentry.prod.mozaws.net/operations/basket-prod/issues/5947191/).

jgmize:
#195

### DIFF
--- a/basket/news/backends/sfdc.py
+++ b/basket/news/backends/sfdc.py
@@ -138,7 +138,7 @@ def to_vendor(data):
             data['lang'] = 'en'
 
     for k, v in data.items():
-        if v != '' and k in FIELD_MAP:
+        if v is not None and v != '' and k in FIELD_MAP:
             if k in PROCESSORS_TO_VENDOR:
                 v = PROCESSORS_TO_VENDOR[k](v)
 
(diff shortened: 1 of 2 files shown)


metadave:
+1, are there any tests that cover this fix?

pmac:
we should add a test for this issue. I'll add one to this branch quickly and then merge. thanks!

pmac:
@jgmize I added tests, and as part of that changed how it was fixed. Based on how we handle blank values I think those keys with None values should be filtered out entirely, so I've done it that way instead. Let me know what you think.

jgmize:
Thanks @pmac, this is much better than my quick fix was.

PR #200, opened 2019-06-27
Upgrade Django to 2.2.3

### LINKED ISSUE #97: Upgrade Django to 2.2
~~This comes with a need to upgrade Python to 3.6 and Celery to latest (4.2 right now) after Celery is upgraded in #175.~~

Ready!

pmac:
Fix #97


pmac:
Seems very appropriate for #200 to be the upgrade to Django 2.x 🎉

basket/news/views.py
```diff
@@ -228,6 +228,7 @@ def fxa_callback(request):
 @require_POST
 @csrf_exempt
 def confirm(request, token):
+    token = str(token)
```
pmac: These changes are due to the [`<uuid:token>` URL components](https://docs.djangoproject.com/en/2.2/topics/http/urls/#path-converters) returning [`UUID` objects](https://docs.python.org/3/library/uuid.html#uuid.UUID) instead of strings.

pmac:
Updated to Django 2.2.3

jpetto:
Thinking we should merge this tomorrow morning in case it might impact the testing going on on dev.

PR #205, opened 2019-07-11
Fix email missing in common-voice task retries

pmac:
The task was modifying the data dict passed to it (removed the email key). This meant that the data passed to the task on a retry due to a failure of the SFDC call would be missing the email and would fail with a KeyError. This change fixes that and adds tests for the task.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ jobs:
       - run:
           name: Run Tests
           command: |
-            bin/dc.sh run --rm --name test-run test-image
+            bin/dc.sh run --name test-run test-image
       - run: docker cp test-run:/app/test-results $TEST_RESULTS
       - store_test_results:
           path: /tmp/test-results
--- a/basket/news/tests/test_tasks.py
+++ b/basket/news/tests/test_tasks.py
@@ -29,6 +29,7 @@ from basket.news.tasks import (
     process_donation,
     process_donation_event,
     process_petition_signature,
+    record_common_voice_goals,
     RECOVERY_MESSAGE_ID,
     SUBSCRIBE,
     send_recovery_message_task,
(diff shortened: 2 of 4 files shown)


.circleci/config.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       - run:
           name: Run Tests
           command: |
-            bin/dc.sh run --rm --name test-run test-image
+            bin/dc.sh run --name test-run test-image
```
jpetto: Why do we want to stop removing the container here? Does data need to persist between test runs?

basket/news/views.py
```diff
@@ -398,13 +398,6 @@ def common_voice_goals(request):
     if form.is_valid():
         # don't send empty values and use ISO formatted date strings
         data = {k: v for k, v in form.cleaned_data.items() if not (v == '' or v is None)}
-        if 'email' not in data:
-            sentry_client.capture('raven.events.Message', request=request)
-            return HttpResponseJSON({
-                'status': 'error',
-                'errors': 'no valid email address'
-            }, 400)
```
jpetto: I'm probably missing something, but why are we removing this check?

.circleci/config.yml
pmac: The next thing that circle does is to `docker cp test-run:/app/test-results $TEST_RESULTS` which requires that the `test-run` container exist. This step gives more info for the test run report in circle. That command was failing due to the missing container causing the test runs to appear to fail.

basket/news/views.py
pmac: This was mostly for debugging. I was trying to get more info on what I thought were calls to the view that were missing emails. This shouldn't have been possible because the email field in the form above this section has email as a required field. When I figured out that the problems were the retries and due to the data dict modification this was no longer necessary and would never be called.

.circleci/config.yml
jpetto: Ah ha! Noted.

pmac:
Thanks Jon!

basket/news/views.py
jpetto: Makes sense!

basket/news/tests/test_tasks.py
```diff
@@ -1019,3 +1020,49 @@ def test_basetime_provided(self):
...
+        data = {
+            'email': 'dude@example.com',
+            'first_contribution_date': '2018-06-27T14:56:58Z',
+            'last_active_date': '2019-07-11T10:28:32Z',
+            'two_day_streak': False,
+        }
+        record_common_voice_goals(data)
+        # ensure passed in dict was not modified in place.
+        # if it is modified a retry will use the modified dict.
+        assert data == data
```
pmac: oh wow. you and I both missed this. now I feel dumb. this isn't testing anything 🤦‍♂ 

I fixed it in 4d0668d4e5cb304e68b04bcc0d2b278d27b591e6

jpetto: lol. my mind knew the intent! 🤦‍♂

PR #253, opened 2019-07-22
Update subhub field mappings & add a bit of documentation.

jpetto:


basket/news/tasks.py
```diff
@@ -986,7 +1002,7 @@ def process_subhub_event_payment_failed(data):
         transaction_data = {
             'Donation_Contact__c': user_data['id'],
             'PMT_Subscription_ID__c': data['subscription_id'],
-            'Invoice_Number__c': data['number'],
+            'Invoice_Number__c': data['invoice_number'],
```
jgmize: Do we know that `invoice_number` is in data here? In https://github.com/mozmeao/basket/pull/253/files#diff-239e0296d7bbbb9699bc1a16569f11b7R862 we had to check for it explicitly.

jpetto: we are reasonably sure, yes. in the `process_subhub_event_subscription_charge` method, we are handling two different payloads, which is why we have to check.

PR #254, opened 2019-07-22
Update monitored flag for subub cancellations.

jpetto:
intent is to branch off truthy/falsy value contained in `cancel_at_period_end`.

sample incoming data looks like:

```
data = {
        "event_id": "evt_00000000000000",
        "event_type": "customer.subscription.updated",
        "customer_id": "cus_00000000000000",
        "subscription_id": "sub_00000000000000",
        "subscriptionId": "sub_00000000000000",
        "plan_amount": 500,
        "canceled_at": 1519680008,
        "cancel_at": 1521854209,
        "cancel_at_period_end": False,
        "nickname": "subhub",
        "active": False,
        "productName": "subhub",
        "invoice_id": "in_test123",
        "created": 1519363457,
        "amount_paid": 500,
        "eventId": "evt_00000000000000",
        "messageCreatedAt": int(time.time()),
        "eventCreatedAt": 1326853478,
        "uid": "user123",
}
```

### DIFF
--- a/basket/news/tasks.py
+++ b/basket/news/tasks.py
@@ -801,7 +801,7 @@ def process_subhub_event_customer_updated(data):
 @et_task
 def process_subhub_event_subscription_charge(data):
     """
-    Event names: invoice.finalized & payment_intent.succeeded
+    Event names: invoice.finalized, payment_intent.succeeded
 
     This method handles both new and recurring charges.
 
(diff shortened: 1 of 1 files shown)


basket/news/tasks.py
```diff
@@ -929,7 +929,7 @@ def process_subhub_event_subscription_updated(data):
         return
 
     # it's only a cancellation if cancel_at is populated
```
pmac: should also update this comment

PR #265, opened 2019-07-29
Add a short TTL to the Oauth token request to FxA

### LINKED ISSUE #256: Reduce the TTL of the OAuth access token for FxA
See https://github.com/mozilla/fxa/issues/1810

pmac:
Fix #256


pmac:
Updated to fix test failures.

PR #266, opened 2019-07-29
[fix #263] Do math on amount for subscription created/updated.

### LINKED ISSUE #263: Close Date & Amount on New Subscriptions / SubHub
When new subscriptions are coming in to Salesforce (SFDC), the amount is incorrect and the Close Date is defaulting to 1/19/1970. The amount calculation is $500.00 and should be $5.00 (this is likely a math issue) - screen shot of data is included below. 

Fields in SFDC are: Opportunity.CloseDate and Opportunity.Amount

The events that would trigger from a new subscription are: customer.subscription.created, invoice.finalized and payment.intent.succeeded 

The last 2 events (invoice.finalized and payment.intent.succeeded are working as expected in other flows, so we may want to check customer.subscription.created first to see if the issue is in that event handling logic)

jpetto:
Also refactor to be a bit more DRY.

### DIFF
--- a/basket/news/tasks.py
+++ b/basket/news/tasks.py
@@ -931,39 +931,36 @@ def process_subhub_event_subscription_updated(data):
     if not user_data:
         return
 
+    # just in case amount data is missing/malformed
+    try:
+        # amount values are cents - convert to dollars
+        plan_amount = int(data['plan_amount']) / float(100)
+    except ValueError:
+        plan_amount = 0
+
+    # common transaction data for all three scenarios
+    transaction_data = {
+        'Amount': plan_amount,
+        'PMT_Subscription_ID__c': data['subscription_id'],
+        'Service_Plan__c': data['nickname'],
+    }
+
     # it's only a cancellation if cancel_at_period_end is truthy
     if (data['cancel_at_period_end']):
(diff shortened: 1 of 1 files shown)


basket/news/tasks.py
```diff
@@ -931,39 +931,36 @@ def process_subhub_event_subscription_updated(data):
...
-            'CloseDate': iso_format_unix_timestamp(data['canceled_at']),
-            'Donation_Contact__c': user_data['id'],
-            'Name': 'Subscription Services',
-            'PMT_Subscription_ID__c': data['subscription_id'],
-            'Payment_Source__c': 'Stripe',
-            'RecordTypeId': settings.SUBHUB_OPP_RECORD_TYPE,
-            'StageName': 'Subscription Canceled',
-            'Service_Plan__c': data['nickname'],
-        }
+        transaction_data['Billing_Cycle_End__c'] = iso_format_unix_timestamp(data['cancel_at'])
```
pmac: style NIT: If you'd rather keep all of this as a dict literal instead of repeating `transaction_data` so much you can always `transaction_data.update({...})` instead of adding them all separately. No need to rewrite, just FYI as it could have saved you some typing.

PR #269, opened 2019-07-30
[fix #268] Refactors subhub initial purchase flag logic.

### LINKED ISSUE #268: Initial Purchase flag - Subscription Services
Please set Initial_Purchase__c on the Opportunity to TRUE on the code that processes the payload from customer.subscription.created - that will denote if the charge is for a new subscription or a recurring charge. 

Any code that checked for Opportunity with the same subscription ID should be removed from existing code as it will not work with multiple data payloads coming in from SubHub.

jpetto:


basket/news/tasks.py
```diff
@@ -966,6 +947,8 @@ def process_subhub_event_subscription_updated(data):
             'StageName': 'Closed Won',
         })
 
+        transaction_data['Initial_Purchase__c'] = True if data['event_type'] == 'customer.subscription.created' else False
```
pmac: can just be this:
```suggestion
        transaction_data['Initial_Purchase__c'] = data['event_type'] == 'customer.subscription.created'
```

jpetto: of course it can look just like JS 😀

PR #283, opened 2019-08-08
Subhub - Removes invoice_id from cancellation opportunities.

jpetto:


pmac:
It technically looks fine. I'm just confused why it's a problem to send the TX ID with the update.

jpetto:
no, you're right. there's no issue sending it with the update - only with the cancellation. will update.

PR #296, opened 2019-08-15
Configure celery for SQS

pmac:
Re https://github.com/mozmeao/infra/issues/1149


pmac:
I think we're not going to do this after all.

PR #305, opened 2019-08-19
Amo contacts sync

### LINKED ISSUE #223: Create AMO Contact Sync API Endpoint
Create a URL that accepts JSON-formatted data POST requests and requires a valid API key. See #148 for information on expected data and SFDC field names. We'll need to decide on the JSON schema that we'll expect from AMO.

pmac:
Update AMO users and addons in SFDC as sent to us by AMO.

Fix #223


pmac:
Still a WIP but worth taking an initial look at when you get a chance @jpetto

basket/news/backends/sfdc.py
```diff
@@ -309,27 +317,45 @@ def campaign_member(self):
...
     @time_request
-    def get(self, token=None, email=None, payee_id=None):
+    def get(self, token=None, email=None, payee_id=None, amo_id=None):
         """
         Get a contact record.
 
         @param token: external ID
         @param email: email address
         @param payee_id: external ID from Stripe/payment processor
+        @param payee_id: external ID from AMO
```
jpetto: little bit o' copy pasta here? s/payee_id/amo_id?

basket/news/backends/sfdc.py
```diff
@@ -309,27 +317,45 @@ def campaign_member(self):
...
         @return: dict
         """
-        assert token or email or payee_id, 'token, email, or payee_id is required'
+        assert token or email or payee_id or amo_id, 'token, email, or payee_id is required'
```
jpetto: should update the error message to include `amo_id`?

basket/news/backends/sfdc.py
```diff
@@ -345,7 +371,8 @@ def add(self, data):
         """
         if not data.get('last_name', '').strip():
             data['last_name'] = LAST_NAME_DEFAULT_VALUE
-        self.contact.create(to_vendor(data))
+
+        return self.contact.create(to_vendor(data))
```
jpetto: ah, this was the line you were mentioning earlier that returns the id, right? nice.

basket/news/tasks.py
```diff
@@ -879,7 +880,7 @@ def process_subhub_event_subscription_charge(data):
             # if the following succeeds, we don't have to worry about setting
             # the initial purchase flag - we leave it alone, assuming it was
             # set properly in the except block below
-            sfdc.opportunity.update('PMT_Transaction_ID__c/{}'.format(charge_id), transaction_data)
+            sfdc.opportunity.update(f'PMT_Transaction_ID__c/{charge_id}', transaction_data)
```
jpetto: noting that this will prob throw a merge conflict now that the API has been majorly refactored. (we are no longer doing any updates, so either way can revert this.)

basket/news/tasks.py
```diff
@@ -1252,6 +1251,87 @@ def process_petition_signature(data):
             raise
 
 
+def upsert_amo_user_data(data):
+    """
+    Update AMO user data in the SFDC contact, or create a contact.
+    Return the Contact data (the contact ID at a minimum).
+
+    :param amo_data: dict of amo user data
```
jpetto: nit - param is just `data`

basket/news/tasks.py
```diff
@@ -1252,6 +1251,87 @@ def process_petition_signature(data):
...
+    user = sfdc.add(amo_data)
+    # need amo_id for linking addons and authors
+    user['amo_id'] = amo_id
+    return user
+
+
+def amo_compress_categories(categories):
+    cats_list = []
+    for product, cats in categories.items():
+        cats_list.extend([f'{product}-{cat}' for cat in cats])
```
jpetto: 🐈

basket/news/backends/sfdc.py
pmac: yup. simple as that 🤦‍♂

basket/news/tasks.py
pmac: yeah. I got a bit carried away with f-strings. I'll revert this. But yay f-strings 🎉

pmac:
@jpetto I believe I've fixed everything up and have now added tests for the addons task as well. Should be ready to go.

basket/news/tasks.py
```diff
@@ -1252,6 +1251,86 @@ def process_petition_signature(data):
...
+        'AMO_Slug__c': data['slug'],
+        'AMO_Status__c': data['status'],
+        'AMO_Type__c': data['type'],
+        'AMO_Update__c': data['last_updated'],
+        'Average_Daily_Users__c': data['average_daily_users'],
+        'Dev_Disabled__c': 'Yes' if data['is_disabled'] else 'No',
+        'Name': data['name'],
+    }
+    sfdc.addon.upsert(f'AMO_AddOn_Id__c/{data["id"]}', addon_data)
+    addon_record = sfdc.addon.get_by_custom_id('AMO_AddOn_Id__c', data['id'])
```
jpetto: do you think we need any checks around this returning valid data?

pmac: My thinking was that we just added it, so it should be there. If it fails to fetch anything it should raise an exception and the task should retry. It's unfortunate we have to do this but it's because the `upsert()` function does not return an ID for the object it just created or updated.

PR #310, opened 2019-08-21
Add tests for SubHub API tasks & views

jpetto:


basket/news/tests/test_tasks.py
```diff
@@ -286,19 +310,173 @@ def test_credit_card_expiring(self, sfmc_mock, sfdc_mock, gud_mock):
...
+            'currency': 'us',
+        }
+
+        process_subhub_event_payment_failed(data)
+
+        gud_mock.assert_called_with(payee_id=data['customer_id'], extra_fields=['id'])
+
+        sfdc_mock.opportunity.create.assert_called_with({
+            'Amount': cents_to_dollars(data['amount_due']),
+            'CloseDate': iso_format_unix_timestamp(data['created']),
```
pmac: this is fine, but a better test would be to hard code the expected output of this function and the one above for Amount. This will make sure things stay right even if one of these functions change or we swap it out.

PR #311, opened 2019-08-22
Refactor charges to be upserts (for re-instated plans).

jpetto:


basket/news/tasks.py
```diff
@@ -793,15 +793,17 @@ def process_subhub_event_subscription_charge(data):
...
             'StageName': 'Closed Won',
         }
 
-        sfdc.opportunity.create(transaction_data)
+        # if a customer re-instates service after a cancellation, the record needs to be
+        # updated
+        sfdc.opportunity.upsert(f'PMT_Invoice_ID__c/{data["invoice_id"]}',
+                                transaction_data)
     else:
         statsd.incr('news.tasks.process_subhub_event.subscription_charge.user_not_found')
```
pmac: Unrelated, but should we retry here? If there's no user, perhaps there's just no user yet, and retrying a few times maybe it'll happen during the wait?

PR #313, opened 2019-08-22
Add ability to enable Fernet encryption for task data

### LINKED ISSUE #270: Encrypt task messages in the queue
Using a technique similar to one of [these](https://blog.adimian.com/2018/07/18/securing-celery-and-redis-without-ssl/) [posts](https://blog.heroku.com/securing-celery) we should be able to seamlessly use symmetric encryption to ensure data on the queue is safely unreadable without the key.

The technique involves the use of [Fernet](https://cryptography.readthedocs.io/en/latest/fernet/) from the Python Cryptography library, which is already installed in basket.

pmac:
Encrypts Celery task messages while on the queue.

Fix #270

## For testing setup

Add a Fernet key to your settings in `.env`:

1. get a basket shell `docker-compose run web python manage.py shell`
2. then `from cryptography.fernet import Fernet` 
3. and finally `Fernet.generate_key()`
4. take the result of step 3 and add it to your `.env` file `KOMBU_FERNET_KEY=<key_from_above>`

Run basket: `docker-compose up web worker`

Then have it run a task using e.g. curl: `curl -d "email=pmac-fenet-test@restmail.net" -d "newsletters=firefox" http://localhost:8000/news/subscribe/`

Note: you do have to have at least one newsletter in your basket database with the name from the above curl command in order for that one to work.


pmac:
This is a draft PR for a possible method of encrypting messages on the basket queue. Thoughts @mozmeao/theconn ?

basket/news/celery.py
```diff
@@ -5,11 +5,44 @@
...
 from raven.contrib.django.raven_compat.models import client
 
 
+FERNET = None
+
+if settings.KOMBU_FERNET_KEY:
+    FERNET = Fernet(settings.KOMBU_FERNET_KEY)
+    if settings.KOMBU_FERNET_KEY_PREVIOUS:
+        # this will try both keys. for key rotation.
+        FERNET = MultiFernet([FERNET, Fernet(settings.KOMBU_FERNET_KEY_PREVIOUS)])
```
glogiotatidis: :+1: nice

pmac:
Yeah. It's simplicity is nice, and I like that it will work no matter what celery broker we use. Thanks for taking a look!

pmac:
This is a big-ish change. I've tested locally with the docker-compose setup which uses Redis for the celery broker. What say you @mozmeao/theconn ?

basket/news/celery.py
```diff
@@ -5,11 +5,44 @@
...
+    message = json.dumps(message)
+    return FERNET.encrypt(force_bytes(message))
+
+
+def fernet_loads(encoded_message):
+    try:
+        encoded_message = FERNET.decrypt(force_bytes(encoded_message))
+    except InvalidToken:
+        # likely a message that was not encrypted
+        pass
```
jgmize: Is the thinking here that we would be pulling unencrypted messages from the queue right after deploying this? It seems like we might want some sort of notification, such as incrementing a statsd counter, so that we can confirm that there aren’t any unencrypted messages flowing through the system somehow after a certain point.

pmac: that's a good idea. I've been thinking about this bit and it "shouldn't" ever happen. The way Celery will handle this is that the messages on the queue at deployment won't be encrypted of course, but also they were sent using the normal "json" transport and have a different mime type that is tied to the "json" transport decoder at the other end. So those messages should go through and then stop as the default transport changes at deployment once an encryption key is configured. We may want to just allow the error to happen here so that Sentry will notify us as well as your good suggestion of adding some more instrumentation.

pmac:
To be clear, a lot of this was based on the code in https://github.com/heroku/kombu-fernet-serializers/, but I decided not to directly use that project as it seems to have stagnated and it wasn't using the same JSON machinery as the default kombu JSON stuff. It seemed simple enough to just add to basket.

pmac:
I've rebased and added statsd to the fernet functions.

pmac:
@jgmize I've updated how this works so that our functions are the only ones used by unregistering the default "json" serializer, and registering our own as "json" so that all messages flow through our functions, allowing us to instrument everything.

PR #322, opened 2019-09-06
Increase max retries to 11

### LINKED ISSUE #321: Increase basket's task retry timeout
The request is to have the tasks continue to retry for at least 48 hours.

cc @metadave @andrewmozilla

pmac:
This should result in the last retry happening around 68 hours after the initial attempt using the current algorithm.

Fix #321


pmac:
330px-Spinal_Tap_-_Up_to_Eleven

metadave:
:+1:

PR #349, opened 2019-10-14
Exempt all non-admin URLs from OIDC session refresh

pmac:


pmac:
@jgmize this should do it. Tested locally.

pmac:
Fix #350

PR #352, opened 2019-10-15
Retry all subhub tasks that lookup a user and create/modify an opportunity

pmac:
These were previously simply dropping the data if no user could be found. There are many reasons the get_user_data function might return None though. We should retry all of these.


pmac:
@glogiotatidis I need to go ahead and merge this so that the subhub team can test as soon as possible, but if you wouldn't mind taking a quick look when you can at what changed I'd appreciate it. I'll follow up with a PR if you see anything that needs fixing.

glogiotatidis:
LGTM! :cake:

PR #363, opened 2019-10-24
Add handling for the SubHub customer.subscription.reactivated event

### LINKED ISSUE #362: SubHub > Salesforce integration update to process reactivated subscriptions
Update basket to accept new event type for reactivated subscriptions. Mappings can be found here ("Reactivation" worksheet) (https://docs.google.com/spreadsheets/d/1z6aXj_9GUpl5g_hR2k_SmjY5tyhxvnoHGOr8e-cbo18/edit#gid=962422475

Jackie will let us know when the payload is ready to be tested. 

Desiree will update SFDC's sandbox to include the new stage name

pmac:
Fix #362

### DIFF
--- a/basket/news/tasks.py
+++ b/basket/news/tasks.py
@@ -786,6 +786,10 @@ def process_subhub_event_subscription_charge(data):
         statsd.incr('news.tasks.process_subhub_event.subscription_charge.user_not_found')
         raise RetryTask('Could not find user. Try again.')
 
+    nickname = data['nickname']
+    if isinstance(nickname, list):
+        nickname = nickname[0]
+
     # if a customer re-instates service after a cancellation, the record needs to be updated
     sfdc.opportunity.upsert(f'PMT_Invoice_ID__c/{data["invoice_id"]}', {
         'Amount': cents_to_dollars(data['plan_amount']),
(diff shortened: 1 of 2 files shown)


basket/news/tasks.py
```diff
@@ -805,11 +809,43 @@ def process_subhub_event_subscription_charge(data):
...
+@et_task
+def process_subhub_event_subscription_reactivated(data):
+    statsd.incr('news.tasks.process_subhub_event.subscription_reactivated')
+    user_data = get_user_data(payee_id=data['customer_id'],
+                              extra_fields=['id'])
+    if not user_data:
+        statsd.incr('news.tasks.process_subhub_event.subscription_reactivated.user_not_found')
+        raise RetryTask('Could not find user. Try again.')
+
+    nickname = data['nickname']
```
jgmize: Based on my current understanding a missing key error here would result in a retry loop until the task times out. If we get any invalid events with missing keys, what is the expected/desired behavior for this task?

pmac: No, it would simply throw the exception and be recorded in Sentry and fail. Only specific exception types are retried:

https://github.com/mozmeao/basket/blob/e5fbaa77b9f12b87a5e4b8aecc33c09dd88d01bb/basket/news/tasks.py#L174-L177

pmac:
Thanks! I have a couple of questions about the expected data outstanding in the goog sheet from the issue. I'll merge after those are sorted.

pmac:
Based on the SubHub PR I think this is now good to go. I'll get it merged shortly and we can test on dev.

https://github.com/mozilla/subhub/pull/330/

PR #376, opened 2019-11-19
Add a Makefile for easier local CLI use

pmac:
Also use make commands in CI


bin/run-tests.sh
```diff
@@ -4,5 +4,6 @@ set -exo pipefail
 
 flake8 basket
 urlwait
+python manage.py makemigrations | grep "No changes detected"
```
glogiotatidis: :+1:

PR #396, opened 2019-12-13
SubHub: Look up new users by FxA_ID first to avoid dupes

### LINKED ISSUE #395: Lookup new SubHub users by FxA ID first
We currently encounter errors when a user comes in via SubHub with an email address and FxA ID, but another contact in Salesforce already has that FxA ID. This can happen due to the user changing their FxA email address at some point. To combat this @mozbrandon, @andrewmozilla, and I discussed solutions and decided on the following:

Look up users by FxA ID first. If a record is found:

* If the email matches the one FxA sent us, then we've found the right record and can continue on.
* If the emails don't match then we update the contact found above to have the `FxA_Id__c` field contain `DUPE:<fxa_id>` and set the `FxA_Account_Deleted__c` field to `True`. We then need to look up the FxA email to see if there's a contact for that email.
  * If there is a contact for the FxA email then fetch it and update it to have the correct FxA ID.
  * If there is not a contact for the FxA email then create one

I think that's it. @andrewmozilla please correct me if I'm wrong.

pmac:
If another contact already has the FxA_ID but a different email address we get errors. This should solve that.

Fix #395

### DIFF
--- a/basket/news/tasks.py
+++ b/basket/news/tasks.py
@@ -741,10 +741,29 @@ def process_subhub_event_customer_created(data):
         'payee_id': data['customer_id']
     }
 
-    user_data = get_user_data(email=data['email'])
+    user_data = None
+    # try getting user data with the fxa_id first
+    user_data_fxa = get_user_data(fxa_id=contact_data['fxa_id'],
+                                  extra_fields=['id'])
+    if user_data_fxa:
+        # if the email matches what we got from subhub, which got it from fxa, we're good
+        if user_data_fxa['email'] == data['email']:
+            user_data = user_data_fxa
+        # otherwise we've gotta make sure this one doesn't interfere with us updating or creating
+        # the one with the right email address below
+        else:
+            statsd.incr('news.tasks.process_subhub_event.customer_created.fxa_id_dupe')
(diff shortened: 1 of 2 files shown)


basket/news/tasks.py
```diff
@@ -741,10 +741,29 @@ def process_subhub_event_customer_created(data):
         'payee_id': data['customer_id']
     }
 
-    user_data = get_user_data(email=data['email'])
+    user_data = None
+    # try getting user data with the fxa_id first
+    user_data_fxa = get_user_data(fxa_id=contact_data['fxa_id'],
+                                  extra_fields=['id'])
```
glogiotatidis: nit: is this `id` from `extra_fields` actually used? same for line 763.

pmac: I've been thinking about what to do with that. It's annoying to have to include that in every call. It's there because I need it to return the `id` field so that when I pass the `user_data` dict to `sfdc.update()` it'll have the ID in there to use to update the record. `id` is excluded normally so that we're sure to avoid sending it out with the other user data in HTTP responses, but it can be filtered at that point.

PR #402, opened 2019-12-18
Convert proration_amount from cents to dollars in SubHub subscription_updated task

### LINKED ISSUE #401: Proration_Amount__c Decimal Location
The decimal location of `Proration_Amount__c` is off by two marks.
Example: Payload value of 356 resolves to $356.00 in Salesforce; the expected result is $3.56.

pmac:
This one was missed during implementation.

Fix #401


pmac:
Minimal change and the tests pass. Going ahead and merging.

PR #404, opened 2019-12-19
Record FxA Verified events in SFDC

### LINKED ISSUE #149: Put FxA user data into SFDC in addition to SFMC
FxA users will soon need to go into SFDC instead of the data extension in SFMC into which they're going now. The timing for this will need to be coordinated with LCM and the IT Salesforce folks.

pmac:
Also continue to record them in SFMC.

Fix #149


pmac:
@andrewmozilla I've added a switch in the code so that we can go ahead to prod (hopefully tomorrow) and then only enable the flow of data to SFDC when we're ready.

basket/news/tasks.py
```diff
@@ -1350,3 +1346,34 @@ def snitch(start_time=None):
     requests.post('https://nosnch.in/{}'.format(snitch_id), data={
         'm': totalms,
     })
+
+
+def get_fxa_user_data(fxa_id, email):
```
glogiotatidis: Does it make sense to use this function here too?

https://github.com/mozmeao/basket/blob/a1df42d257d0b8bb63c41882eabd05b24adf7fe3/basket/news/tasks.py#L749

basket/news/tasks.py
```diff
@@ -741,28 +758,7 @@ def process_subhub_event_customer_created(data):
...
-            sfdc.update(user_data_fxa, {
-                'fxa_id': f"DUPE:{contact_data['fxa_id']}",
-                'fxa_deleted': True,
-            })
-
-    # if we still don't have user data try again with email this time
-    if not user_data:
-        user_data = get_user_data(email=data['email'], extra_fields=['id'])
-
+    user_data = get_fxa_user_data(data['user_id'], data['email'])
```
pmac: @glogiotatidis  I did change that out here (re: your comment on DRY)

pmac:
Thanks for the quick review as always @glogiotatidis 🥇

PR #409, opened 2020-01-06
SubHub: Fix errors when nickname is an empty list

pmac:


pmac:
See https://sentry.prod.mozaws.net/operations/basket-prod/issues/7009030/

PR #410, opened 2020-01-07
Add more contact data for FxA Verified events in SFDC

### LINKED ISSUE #405: Updates to FxA Feed to SalesCloudcCRM
This request is related to the work that has already been done on the Basket side around turning on the feed for FxA Contacts to come into Salesforce CRM. 

UPDATES FROM LCM:
For Net Net New verified account holder records feeding into CRM:

- Bring over FxA ID
- Flip Double Opt In to True
- Flip Subscriber to True
- Flip Email Format to H
- Flip Firefox Account Tips Subscriber field to True
- Flip any other applicable subscription fields to True that were a part of their signup process (i.e. FxA subscription options upon signup: Knowledge is Power Subscription, New Product Testing Subscriber, Take Action for the Internet Subscription)
- Bring over FxA language and populate Newsletter Language (based on previously discussed & agreed upon mapping - one of our standard 9 locales or “Other”; I believe you wanted to capture the ACTUAL language from FxA in another new field too in case it was ever needed.)

pmac:
Fix #405

### DIFF
--- a/basket/news/tasks.py
+++ b/basket/news/tasks.py
@@ -276,7 +276,6 @@ def fxa_verified_sfdc(data):
     fxa_id = data['uid']
     lang = data['lang']
     create_date = data.get('createDate')
-    subscribe = data.get('marketingOptIn')
     newsletters = data.get('newsletters')
     metrics = data.get('metricsContext', {})
     new_data = {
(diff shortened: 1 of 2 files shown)


basket/news/tasks.py
```diff
@@ -286,21 +285,15 @@ def fxa_verified_sfdc(data):
...
-            newsletters.append(settings.FXA_REGISTER_NEWSLETTER)
-
-        add_news = ','.join(newsletters)
-    elif subscribe:
-        add_news = settings.FXA_REGISTER_NEWSLETTER
-
-    if add_news:
-        new_data['newsletters'] = add_news
+    newsletters = newsletters or []
+    newsletters.append(settings.FXA_REGISTER_NEWSLETTER)
```
glogiotatidis: Just checking that it's OK to `append` if `FXA_REGISTER_NEWSLETTER` already exists since you have a check for this in previous revision. https://github.com/mozmeao/basket/pull/410/files#diff-239e0296d7bbbb9699bc1a16569f11b7L295

pmac: Good catch. I did track down the code path after this and it does set math on the list, so it gets uniqueified later and so it didn't seem worth it to ensure uniqueness here as well.

PR #476, opened 2020-02-24
For FxA Verified events use "ts" value if "createDate" not present

### LINKED ISSUE #472: Update FxA Created Date in SFMC based on timestamp data in verified events
In #77 we added support for a `createDate` from FxA. They no longer send that field with `verified` events. There are a few other fields that do come with the events that could potentially work. For example, the data looks like this:

```python
{
  'country': 'United States', 
  'countryCode': 'US', 
  'email': 'sync0.3486440592016071@restmail.net', 
  'event': 'verified',
  'iss': 'api-accounts.stage.mozaws.net', 
  'lang': 'en', 
  'locale': 'en', 
  'metricsContext': {
    'flow_time': None, 
    'time': '** ** ** **'
  }, 
  'timestamp': '** ** ** **', 
  'ts': 1582252410.066, 
  'uid': 'dac5bf77929745668ca44ea392e35a4c', 
  'userAgent': 'node-XMLHttpRequest'
}
```

So it looks like `metricsContext['time']`, `timestamp`, or `ts` values could work. I've asked in https://github.com/mozilla/fxa/issues/4138 which would be the best and most reliable one to use. Once we find out we should be able to finish #125.

cc @desiree-mofo

pmac:
Fix #472

### DIFF
--- a/basket/news/tasks.py
+++ b/basket/news/tasks.py
@@ -276,7 +276,7 @@ def fxa_verified_sfdc(data):
     email = data['email']
     fxa_id = data['uid']
     lang = data['lang']
-    create_date = data.get('createDate')
+    create_date = data.get('createDate', data.get('ts'))
     newsletters = data.get('newsletters')
     metrics = data.get('metricsContext', {})
     new_data = {


pmac:
@desiree-mofo this will use the event creation date for the `FxA_Created_Date__c` field. That seems like our best option for now.

PR #480, opened 2020-02-27
Introduce jitter into the exponential backoff retry delays

pmac:


basket/news/tasks.py
```diff
@@ -64,6 +65,26 @@
 ]
 
 
+def exponential_backoff(retries):
+    """
+    Return a number of seconds to delay the next task attempt using
+    an exponential back-off algorithm with jitter.
+
+    See https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
```
glogiotatidis: very interesting :eyes:

basket/settings.py
```diff
@@ -244,10 +244,11 @@ def path(*args):
 KOMBU_FERNET_KEY_PREVIOUS = config('KOMBU_FERNET_KEY_PREVIOUS', None)
 CELERY_TASK_ALWAYS_EAGER = config('CELERY_TASK_ALWAYS_EAGER', DEBUG, cast=bool)
 CELERY_TASK_SERIALIZER = 'json'
-CELERY_TASK_ACKS_LATE = config('CELERY_TASK_ACKS_LATE', False, cast=bool)
+CELERY_TASK_ACKS_LATE = config('CELERY_TASK_ACKS_LATE', True, cast=bool)
```
glogiotatidis: According to Celery docs, this means that the task may be multiple times executed. I believe this is alright from a data standpoint but maybe it reduces our SF api quota faster?

pmac: That's true. In theory this should only apply if the task doesn't complete before the worker is killed. So this should ensure that the tasks are executed fully at least once even if the pod or node is suddenly destroyed. In theory it should not mean much duplication. I'll definitely be keeping an eye on it thought. Good call-out.

pmac:
Thanks for the review!

PR #530, opened 2020-05-12
[Fix #522] Update AMO sync.

### LINKED ISSUE #522: AMO <> Salesforce Expanded Integration Phase 1
Here's what's needed for phase 1 of the expanded amo integration. 

**Field Mapping:** https://docs.google.com/spreadsheets/d/1ejdszPXF7WIX284UTpXM9Fs2PonTRycGQwEtrkkRT2A/edit?usp=sharing

**Action/Outcome:** https://docs.google.com/document/d/1PIBgZmxbfI-LZlSkB0j4N3wUeJFb2v1JJos8QPHnNSk/edit?usp=sharing

Thank you!

glogiotatidis:


basket/news/tasks.py
```diff
@@ -1335,13 +1336,14 @@ def upsert_amo_user_data(data):
...
     amo_data['amo_id'] = amo_id
-    user = get_user_data(email=email, extra_fields=['id'])
+    user = get_user_data(fxa_id=fxa_id, extra_fields=['id'])
     if user:
         sfdc.update(user, amo_data)
         # need amo_id for linking addons and authors
         user['amo_id'] = amo_id
         return user
 
+    # User not found in database
```
glogiotatidis, line 1346: This the section below for the case user is not found in the database with either `amo_id` or `fxa_id`? 

Could this ever be true, given that AMO users sign in with FxA now? If yes maybe we should request in https://github.com/mozilla/addons-server/issues/14318 to send both FxA ID *and* email so we can create the record @pmac?

pmac: It shouldn't ever be true, but it's possible some user wasn't synced properly with SFDC so I think we have to deal with the possibility. We could just ignore these, but that also seems suboptimal. Maybe we should ask for email. I think I remember us discussing this and deciding against it, but I don't recall why. @diox do you remember?

diox: Well the idea was that FxA is always the source of truth - we shouldn't need to store the email elsewhere. It seems even more suboptimal to have another field...

pmac: It wouldn't mean another field. What we're wondering is if we should create a record in the CRM in the case that we find no record for the `fxa_id` we receive for you. It's possibly that the sync from FxA wasn't perfect. This should almost never happen. But if it did and you also sent us email then we could create a record for the user. If we have no email and we can't find the FxA ID then we have to just ignore the update.

diox: Yeah, still not sure I like that very much. It would hide the problem, but I'm concerned about the possibility of opening up new edge cases...

glogiotatidis: Updated code to ignore this situation, since we don't have the email address. We can increase a statsd counter and/or log an error to Sentry if we see value.

glogiotatidis:
Holding off tests until we agree on the logic

basket/news/tasks.py
```diff
@@ -1322,35 +1322,42 @@ def upsert_amo_user_data(data):
...
+            # another AMO account so we can leave the link.
+            amo_data['amo_deleted'] = amo_deleted
+    else:
+        if fxa_id is None:
+            # User takeover protection, we don't want do keep linking that AMO
+            # account with that FxA account in Salesforce
+            amo_data['amo_id'] = None
+
+    sfdc.update(user, amo_data)
+    return user
```
pmac, line 1360: flake8 is complaining that there needs to be another empty line here.

Issue #532, opened 2020-05-18
Store FxA Primary Email Change in SFDC


pmac:
We currently store these events in SFMC in a data extension, but we should store the new email in the CRM. We've decided to create a new field (e.g. `FxA_Primary_Email__c`) for the changed email. This is related to #522 since AMO uses FxA for auth and gets their email addresses from FxA. 

After this is done we'll need to load the past email changes stored in the data extension in SFMC into the CRM.

/cc @andrewmozilla

glogiotatidis:
Adding the new `FxA_Primary_Email__c` field means that we keep the existing `email` as is and we only update this new field? In other words, why do we need a new field since we already store emails?

pmac:
This is for when people change their primary email in FxA, but if we change their primary contact email then everything would change, and we don't inform them of that yet. So this is a way to let us use their primary FxA email for FxA stuff and the original one for other things (e.g. newsletters, donations, etc.). It's just how the email folks want to handle it for now.

glogiotatidis:
Code is merged, IIRC we left this open to migrate the pre-existing FxA SFMC entries to SFDC?

pmac:
No, I think we can close this. Thanks for the reminder.

PR #535, opened 2020-05-27
[Issue #532] Store FxA email changes in SFDC.

glogiotatidis:


basket/news/tasks.py
```diff
@@ -256,11 +256,17 @@ def fxa_email_changed(data):
         # message older than our last update for this UID
         return
 
+    cache.set(cache_key, ts, 7200)  # 2 hr
```
pmac, line 259: I think we should set this cache after updating everything so that if it fails and is retried that it'll try again.

basket/news/tasks.py
```diff
@@ -256,11 +256,17 @@ def fxa_email_changed(data):
...
+
     sfmc.upsert_row('FXA_EmailUpdated', {
         'FXA_ID': fxa_id,
         'NewEmailAddress': email,
     })
-    cache.set(cache_key, ts, 7200)  # 2 hr
+
+    # Update SFDC
+    user_data = get_user_data(fxa_id=fxa_id, extra_fields=['id'])
```
pmac, line 267: And I think we should do the SFDC update first because the SFMC one just inserts rows, so if this task is retried due to an SFDC issue there will be dupes in SFMC. Not a big deal though.

basket/news/tasks.py
```diff
@@ -256,10 +256,25 @@ def fxa_email_changed(data):
...
 
+    # Update SFDC
+    user_data = get_user_data(fxa_id=fxa_id, extra_fields=['id'])
+    if user_data:
+        sfdc.update(user_data, {'fxa_primary_email': email})
+    else:
+        # FxA record not found, try email
+        user_data = get_user_data(email=email, extra_fields=['id'])
+        if user_data:
+            sfdc.update(user_data, {'fxa_id': fxa_id})
```
glogiotatidis, line 267: Hey @mozbrandon @andrewmozilla can you please advise on whether we should also set `FxA_Primary_Email__c` in this case, i.e. when we find the user in SFDC only by `email` and they have no `FxA ID` assigned.

basket/news/tasks.py
```diff
@@ -256,10 +256,25 @@ def fxa_email_changed(data):
...
+        else:
+            # No matching record for Email or FxA ID. Create one.
+            sfdc.add({'email': email, 'fxa_id': fxa_id)
```
glogiotatidis, line 270: Similar question here: If we don't find the user with `email` or `fxa_id` we'll go ahead and create an entry. Do you want to  set `FxA_Primary_Email__c` in addition to `email`?
/cc @mozbrandon @andrewmozilla

mozbrandon: Yes, please set both. Thanks

basket/news/tasks.py
mozbrandon: Yes, please set `FxA_Primary_Email__c` in this case.

Issue #564, opened 2020-07-10
Only synchronize AMO developers to basket


diox:
We only want AMO data in Salesforce for developers, not all AMO users.

- When an add-on sync is triggered, existing authors info for that add-on in Salesforce needs to be checked to remove authors that are not part of the new sync data: those users are no longer authors of the add-on.
- User sync requests that don't match an user that is already in Salesforce should be ignored. New users should only come from an add-on sync authors data, the user sync should only be used to update existing users.

In addition, we need to handle deletion, so:
- AMO data for users marked as deleted in AMO need to be removed periodically
- Deleted add-ons need to be removed periodically
- AMO data for users with only deleted add-ons or no add-ons left should be removed periodically
- Add-ons with only deleted users left as authors should be removed periodically

pmac:
I've got the first 2 points done. The deletion stuff seems more tricky. 

> AMO data for users marked as deleted in AMO need to be removed periodically

This just means removing the AMO_ID from the user when we get a user update that includes `"deleted": true` right?

> Deleted add-ons need to be removed periodically

I'm not currently looking for deleted for addons. Should I just expect a boolean `"deleted"` field in the addon data like I do for the users?

> AMO data for users with only deleted add-ons or no add-ons left should be removed periodically

I'm guessing this should happen whenever an addon is deleted? I can lookup all of the users and check their other addons. That query might not work very well from basket though. Can you send a request to delete a user that no longer has active users?

> Add-ons with only deleted users left as authors should be removed periodically

Similar to above. It'd be a lot easier for salesforce and basket if you could just let us know which addons to delete.

diox:
> I've got the first 2 points done. The deletion stuff seems more tricky.
> 
> > AMO data for users marked as deleted in AMO need to be removed periodically
> 
> This just means removing the AMO_ID from the user when we get a user update that includes `"deleted": true` right?

Yes

> > Deleted add-ons need to be removed periodically
> 
> I'm not currently looking for deleted for addons. Should I just expect a boolean `"deleted"` field in the addon data like I do for the users?

Add-ons are a bit special in that regard, the field you should look for is `status` with the value `deleted`.

> > AMO data for users with only deleted add-ons or no add-ons left should be removed periodically
> 
> I'm guessing this should happen whenever an addon is deleted? I can lookup all of the users and check their other addons. That query might not work very well from basket though. Can you send a request to delete a user that no longer has active users?

[...]

PR #581, opened 2020-09-02
Support removal of AMO Users from Addons

pmac:
Also avoid marking a user as an AMO user during a User Sync call, and only adding that info to a user during an Addon update.

Re #564


glogiotatidis (approved):
lgtm :beers: . add some tests would be nice since this complex.

PR #655, opened 2021-02-18
acoustic transact

pmac:
- Add Acoustic client and Transact email sending ability
- Add model for Acoustic Transact messages and languages
- Add support for sending all tx emails via Acoustic
- Add separate configs for Acoustic Campaign and Transact Orgs
- Add FxA Activity data to Acoustic


basket/news/backends/acoustic.py
```diff
@@ -0,0 +1,75 @@
...
+
+def xml_tag(tag, value=None, **attrs):
+    xmlt = ElementTree.Element(tag, attrs)
+    if value:
+        xmlt.text = value
+
+    return xmlt
+
+
+def transact_xml(to, campaign_id, fields=None):
```
glogiotatidis, line 20: but why xml dear acoustic? :disappointed:

basket/news/models.py
```diff
@@ -361,6 +360,59 @@ def language_list(self):
...
+
+            try:
+                # failing above, try to get the language prefix
+                message = self.get(message_id=message_id, language__startswith=language)
+            except exception:
+                try:
+                    # failing above, try to get the default language
+                    message = self.get(message_id=message_id, language="en-US")
+                except exception:
+                    # couldn't find a message. give up.
```
glogiotatidis, line 382: Do we want to log or send to Sentry?

basket/news/tasks.py
```diff
@@ -622,8 +638,27 @@ def sfdc_add_update(update_data, user_data=None):
...
 
+@et_task
+def send_acoustic_tx_message(email, vendor_id, fields=None):
+    acoustic_tx.send_mail(email, vendor_id, fields)
+
+
+def send_acoustic_tx_messages(data, message_ids):
+    sent = 0
+    for mid in message_ids:
+        vid = AcousticTxEmailMessage.objects.get_vendor_id(mid, data.get("lang", "en"))
```
glogiotatidis, line 649: Here you use `en` but in https://github.com/mozmeao/basket/pull/655/files#diff-c21115099033cfff1ad2585abb19c8ca144a4cac5b26bd74a60b5c87fadd908eR380 you use `en-US`. Just checking that both are correct?

glogiotatidis (approved):
Just a couple of questions. Good job! :100:

basket/news/tasks.py
pmac: That's a good catch. Should probably use the full `en-US` here too since that's what'll be in the DB and will be the most efficient query.

basket/news/models.py
pmac: hmm... good point. Probably so.

pmac: added sentry capture

PR #668, opened 2021-03-23
Remove FxA file data processor

### LINKED ISSUE #663: Remove FxA log data processor
We've decided to handle this data moving to Acoustic from outside of basket. So the `process_fxa_data.py` file and all associated things can be removed.

pmac:
Fix #663


pmac:
@duallain after this merged we'll need to remove the [relevant deployment](https://github.com/mozmeao/basket-config/blob/master/mozmeao-or/basket-prod/fxaworker-deploy.yaml). Do we need to do a manual k8s operation to remove the deployment before removing the file from the repo?

duallain:
Yeah, we'll have to remove the deploy.  Let's do it after the merge, if that sounds ok to you.  Just to prevent deleting it and then recreating it again.

Issue #674, opened 2021-03-29
Add limited, optional CTMS backend


jwhitlock:
Add the initial backend and integration to [CTMS](https://github.com/mozilla-it/ctms-api). The planned features are:

* Initial CTMS settings, including a flag to enable / disable
* A newsletter subscription written to SFDC is also written to CTMS
* Submitting the preference center writes to SFDC and CTMS
* (maybe) preference center can be optionally sourced from CTMS

jwhitlock:
PR #676 added:

* Initial CTMS settings, including a ``CTMS_ENABLED`` flag
* A ``CTMSSession`` to handle the details of OAuth2 authentication
* A ``CTMSInterface`` to access API endpoints directly
* A partial ``CTMS`` class to mirror the ``SFDC`` class. The only implemented function is ``get``.
* When a record is fetched from SFDC, and it doesn't have CTMS's ``email_id``, attempt to fetch it from CTMS. If this is later written to SFDC, SFDC will now have the correct ``email_id`` as generated by BigQuery.

For the next phase, we'll write parallel records to CTMS. To mirror the SFDC workflows, we'll need:
* ✓ The ``get`` method, which will determine if the record already exists
* An ``add`` method, to create a new record, and generate an ``email_id``. CTMS's ``POST`` or ``PUT`` API would work for this.
* An ``update`` method, to update an existing record. After some reading of Basket code, this _could_ be done with the ``PUT`` API, but would require passing the ``GET`` results in the Basket-formatted data, and would increase the changes of losing data due to synchronous edits. Instead, I'm implementing a ``PATCH`` method for partial updates (https://github.com/mozilla-it/ctms-api/issues/128).

Meanwhile, the first CTMS code is in dev, and triggered an issue. The oauthlib library checks the expiration of the token, and raises a ``TokenExpiredError``. This case can be handled in a similar way to an authentication error (request a new access token). I'm also considering adding refresh token capabilities to CTMS (https://github.com/mozilla-it/ctms-api/issues/132), so that ``requests-oauthlib`` can automatically handle token refreshes.

jwhitlock:
Changes merged and deployed to basket-dev:
* PR #679 added the refresh-token workflow
* PR #680 added the ``add`` and ``update`` methods, and dual-writes for the public APIs
* PR #683 added `api_key` to the ignore list

I think using CTMS as the source for the preference center should wait until the private APIs are converted to talk to CTMS, and we've had some time for integration testing and to shake out the bugs.

[...]

PR #676, opened 2021-03-31
Add initial CTMS API integration: Populate Email_Id__c in SFDC

jwhitlock:
Add the initial CTMS API integration, as part of #674.

This change supports this workflow:

* SFDC data is exported to BigQuery (BQ) periodically. Most records have an empty ``Email_Id_c``, the new CTMS ``email_id``.
* BQ adds an ``email_id``, includes this in exports to Acoustic
* CTMS imports from BQ (work in progress)
* Basket fetches from SFDC, and captures ``Email_Id__c`` as ``email_id``. If it is empty, Basket calls CTMS, using a similar interface. If CTMS has a matching record, then Basket uses CTMS's ``email_id``.
* If Basket writes to SFDC to update the record, it populates ``Email_Id__c``, syncing it with BQ.

New settings are used to configure the CTMS backend:
* ``CTMS_ENV`` - 'prod' or 'stage' to select a known ``CTMS_URL``
* ``CTMS_ENABLED`` - True to enable, False to disable CTMS integration
* ``CTMS_URL`` - Select an alternate CTMS API URL
* ``CTMS_CLIENT_ID`` - CTMS Client ID, should start with "id_"
* ``CTMS_CLIENT_SECRET`` - CTMS Client secret, should start with "secret_"

## Manual testing
For local testing, I put my stage credentials in my ``.env`` file, and added some test data, like:

```sh
$ docker-compose run --rm test python manage.py shell
```
```python
from basket.news.backends.ctms import ctms
email = 'test-20210402@example.com'  # you may need to adjust
ctms.interface.post_to_create({'email': {'primary_email': email}})
user_data = ctms.get(email=email)
```

With ``DEBUG=1``, you'll see the CTMS requests flying by. I don't have SFDC credentials to test the full integration.

## Details

``basket.news.backends.ctms`` provides new classes:

* ``CTMSSession`` - Similar to ``requests.session``, with ``session.get``, ``session.post``, etc. It loads an existing token from Redis, and, if there is no token or there is an authentication failure, fetches a token using ``CTMS_CLIENT_ID`` and ``CTMS_CLIENT_SECRET``. There are tests, but with so much mocking that I've considered removing them.
* ``CTMSInferface`` - Provides a method for each of the CTMS API endpoints. Only ``get_by_alternate_id`` and ``get_by_email_id`` are used in current code.
* ``CTMS`` - Provides ``get``, closely modelled on ``SFDC.get``, to load data by the best provided ID and convert it to the flat basket format. It takes a ``CTMSInterface`` to do the work, or ``None`` to silently do nothing.

New functions:

* ``from_vendor``: Converts CTMS nested dictionaries to a Basket dictionary, similar to ``SFDC.from_vendor``
* ``ctms_session``: Creates a ``CTMSSession`` from Django settings
* ``ctms_interface``: Same, for ``CTMSInterface``

New object:
* ``ctms``: An instance of ``CTMS`` with a ``CTMSInterface`` and ``CTMSSession`` based on Django settings, similar to ``sfdc.sfdc``. If CTMS is disabled, ``ctms.interface`` will be ``None``, and ``ctms.get()`` will always return ``None``.

# Tasks
*  ~Add [request-oauthlib](https://requests-oauthlib.readthedocs.io/en/latest/) package~ - Included by pysilverpop
* [x] Add Django settings
* [x] Add low-level interface for authenticated requests
* [x] Add endpoint-specific methods ( ``GET /ctms``, ``GET /ctms/{email_id}``, etc.)
* [x] Read and write SFDC's ``Email_Id__c`` as ``email_id``
* [x] Add ``CTMS.get`` that works in a similar way to ``SFDC.get``, converting to the basket dictionary format
* [x] In ``get_user_data``, call ``CTMS.get`` to populate ``email_id`` when unset in SFDC
* ~Integrate with newsletter subscriptions (send subscriptions to CTMS, CTMS email_id to Salesforce)~ save for future PR
* ~Integrate with preference center (send subscriptions to CTMS, CTMS email_id to Salesforce)~ save for future PR
* [x] Prep for review

### DIFF
--- a/basket/news/utils.py
+++ b/basket/news/utils.py
@@ -25,6 +25,7 @@ from django_statsd.clients import statsd
 from email_validator import validate_email, EmailNotValidError
 
 from basket.news.backends.common import NewsletterException
+from basket.news.backends.ctms import ctms
 from basket.news.backends.sfdc import sfdc
 from basket.news.models import APIUser, BlockedEmail
 from basket.news.newsletters import (
@@ -315,7 +316,7 @@ def get_user_data(
     extra_fields=None,
     get_fxa=False,
 ):
-    """Return a dictionary of the user's data from Exact Target.
+    """Return a dictionary of the user's data from Salesforce.com (SFDC).
     Look them up by their email or payment processor id (e.g. Stripe) if given,
     otherwise by the token.
 
@@ -329,9 +330,11 @@ def get_user_data(
     If `get_fxa` is True then a boolean field will be including indicating whether they are
     an account holder or not.
 
+    If SFDC does not have the CTMS email_id, try to get it from CTMS.
+
     Review of results:
 
-    None = user completely unknown, no errors talking to ET.
+    None = user completely unknown, no errors talking to SFDC.
 
     otherwise, return value is::
 
@@ -340,6 +343,7 @@ def get_user_data(
         'status':  'error',   # errors talking to ET, see next field
         'desc':  'error message'   # details if status is error
         'email': 'email@address',
+        'email_id': CTMS UUID,
         'format': 'T'|'H',
         'country': country code,
         'lang': language code,
@@ -375,6 +379,25 @@ def get_user_data(
 
     if get_fxa:
         user["has_fxa"] = bool(user.get("fxa_id"))
+
+    if not user.get("email_id"):
+        # Add email_id from CTMS, if CTMS knows about this contact
+        ctms_user = None
+        try:
+            ctms_user = ctms.get(
+                token=token,
+                email=email,
+                sfdc_id=user.get("id"),
+                amo_id=amo_id,
+                fxa_id=fxa_id,
+            )
+        except requests.exceptions.HTTPError as error:
+            if error.response.status_code != 404:
+                sentry_sdk.capture_exception()
+        except RuntimeError:
+            sentry_sdk.capture_exception()
+        if ctms_user and "email_id" in ctms_user:
+            user["email_id"] = ctms_user["email_id"]
 
     user["status"] = "ok"
     return user
--- a/basket/news/backends/ctms.py
+++ b/basket/news/backends/ctms.py
@@ -16,6 +16,90 @@ from basket.news.backends.common import get_timer_decorator
 
 time_request = get_timer_decorator("news.backends.ctms")
 
+# Map CTMS group / name pairs to basket flat format names
+# Missing basket names from SFDC integration:
+#  record_type
+#  postal_code
+#  source_url
+#  fsa_*
+#  cv_*
+#  amo_deleted
+#  payee_id
+#  fxa_last_login
+CTMS_TO_BASKET_NAMES = {
+    "amo": {
+        "add_on_ids": None,
+        "display_name": "amo_display_name",
+        "email_opt_in": None,
+        "language": None,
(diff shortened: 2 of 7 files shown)


requirements.txt
```diff
@@ -519,6 +519,10 @@ requests==2.25.1 \
     --hash=sha256:27973dd4a904a4f13b263a19c866c13b92a39ed1c964655f025f3f8d3d75b804 \
     --hash=sha256:c210084e36a42ae6b9219e00e48287def368a26d03a048ddad7bfee44f75871e \
     # via django-mozilla-product-details, django-synctool, mozilla-django-oidc, pybrowserid, pyfxa, pysilverpop, requests-oauthlib, salesforce-fuelsdk, simple-salesforce
+requests_oauthlib==1.3.0 \
```
pmac, line 522: Isn't this the same lib installed a few lines up? Line 514?

jwhitlock: yep, looks like a bad rebase. I think only the comment should change, will try again

jwhitlock:
I think I'm going for something less ambitious for the first PR. The code is close, but I need some new tests and to mock CTMS calls for some existing tests.

My goal for the first integration is to change ``basket.news.utils.get_user_data`` to support this workflow:

* SFDC data is exported to BigQuery (BQ). ``Email_Id__c`` is empty for most records.
* BQ populates this field as needed when it it exports records for CTMS
* CTMS imports from BQ (work in progress), synced w/ SFDC, minus recent changes
* Basket queries user data from SFDC
* (this change) Basket asks for the related CTMS record if ``email_id`` is unset.
  - If CTMS doesn't know (based on token, email, etc.), nothing changes
  - If CTMS has a record (due to SFDC→BQ import), Basket merges in the ``email_id``
* Basket writes to SFDC, including settings the ``email_id``

The change will be that SFDC accumulates ``Email_Id__c`` fields as records change.  SFDC is still the source of truth, but if we wipe out the CTMS database and re-import, the same emails will get the same ``email_id``. 

It's less ambitious than my original plan, but it does get the bulk of CTMS backend code into basket, and could be in basket-dev soon.

pmac (approved):
I feel like I've failed the review since I've no notes, but this is just beautiful. Thanks for this very thorough and well-commented and well-tested work. ⭐⭐⭐⭐⭐

basket/news/utils.py
```diff
@@ -376,6 +380,25 @@ def get_user_data(
     if get_fxa:
         user["has_fxa"] = bool(user.get("fxa_id"))
 
+    if not user.get("email_id"):
```
pmac, line 383: Oh. I do have one thing. Should we add `and settings.CTMS_ENABLED` here, so that we can push this out whenever we're ready and have it continue to work. Or maybe I'm looking at it wrong and `ctms.get()` will return `None` if CTMS isn't enabled?

pmac: Looking at it more I think my last question is the deal. Is that what you were thinking?

jwhitlock: Yes, you have it right. ``ctms.get()`` quickly returns ``None`` when ``settings.CTMS_ENABLED`` is ``False``, due to ``ctms.interface == None``. So, it should be safe to push this code to prod as long as ``settings.CTMS_ENABLED`` is ``False``.

PR #680, opened 2021-04-12
Add CTMS parallel writes

jwhitlock:
During the transition phase from Salesforce (SFDC) to CTMS, it is (hopefully) safe to write to CTMS when:

* Creating a new contact. The new SFDC contact will have a matching ``Email_Id__c`` to CTMS's ``email_id``, and BigQuery will not generate a new one.
* Updating a contact where SFDC knows the ``email_id``. BigQuery may update or replace the CTMS contact during bulk imports or updates, but the ``email_id`` will be the same.

Details:

* Add CTMS version of `to_vendor`, to convert Basket-formatted data to CTMS. Some SFDC data is ignored by CTMS, and some conversions are not performed.
* Add `patch_by_email_id` to `CTMSInterface`, to support partial updates.
* Add `add` to `CTMS`, to support adding contacts.
* Add `update` to `CTMS`, to support partial updates of contacts, with some CTMS-specific features:
  - If AMO account deletion is detected, reset AMO data with `{'amo': 'DELETED'}`.
  - If unsubscribe all is detected, use `{'newsletters': 'UNSUBSCRIBE'}` for bulk unsubscribe.
  - If subscribing to individual newsletters, apply `source_url` (if set) to new subscriptions 
* Update `upsert_contact`, to send data to CTMS:
  * When adding a new contact, try to add it to CTMS first, then to SFDC with an `email_id`. If CTMS fails, still add to SFDC without an `email_id`.
  * When updating a contact, update it in SFDC, and if it has an `email_id`, update it in CTMS
* Update `sfdc_add_update` with similar logic, but adding to CTMS only after successfully adding to SFDC

Includes tests for the changes, except for ``sfdc_add_update`` which has no existing tests.

### DIFF
--- a/basket/news/backends/ctms.py
+++ b/basket/news/backends/ctms.py
@@ -194,6 +194,7 @@ class CTMSSession:
         return resp
 
     get = partialmethod(request, "GET")
+    patch = partialmethod(request, "PATCH")
     post = partialmethod(request, "POST")
     put = partialmethod(request, "PUT")
 
@@ -310,6 +311,31 @@ class CTMSInterface:
         """
         email_id = data["email"]["email_id"]
         return self.put_by_email_id(email_id, data)
+
+    @time_request
+    def patch_by_email_id(self, email_id, data):
+        """
+        Call PATCH /ctms/{email_id} to partially update a CTMS contact by ID
+
+        To update a data element, send the key and the new value:
+
+          {"email": {"primary_email": "new_email.example.com"}}
+
+        To delete a subgroup, resetting to defaults, send the group value as "DELETE":
+
+          {"amo": DELETE}
+
+        To unsubscribe from all newsletters, set to "UNSUBSCRIBE":
+
+          {"newsletters": "UNSUBSCRIBE"}
+
+        @param email_id: The CTMS email_id of the contact
+        @param data: The contact data to update
+        @return: The updated contact data
+        """
+        resp = self.session.patch(f"/ctms/{email_id}", json=data)
+        resp.raise_for_status()
+        return resp.json()
 
 
 class CTMS:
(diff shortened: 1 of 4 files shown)


jwhitlock:
I've added the additional logic for the CTMS group deletion and bulk unsubscribe features, added some more tests and fixed some issues with ``CTMSInterface``, and adjusted comments. This is ready for review.

Some practical tests in a Django shell, with CTMS configured:

```
email = "test-04122021@example.com"
ctms.add({"email": "test-04122021@example.com"})
user_data = ctms.get(email=email)
update_data = {"country": "usa"}
ctms.update(user_data, update_data)
```

basket/news/tasks.py
```diff
@@ -562,6 +563,16 @@ def upsert_contact(api_call_type, data, user_data):
         if settings.MAINTENANCE_MODE:
             sfdc_add_update.delay(update_data)
         else:
+            ctms_data = update_data.copy()
+            email_id = ctms_data["email_id"] = generate_token()
```
pmac, line 567: Didn't realize basket would be generating CTMS email IDs.

pmac (approved):
I think we're good enough here. We can keep tweaking, but let's get it in for the demo. 🦊

basket/news/tasks.py
jwhitlock: We can go either way, since it is a v4 UUID. This felt easier for the first code, but we should probably pick the general strategy going forward.

For this version, ``POST /ctms`` uses the ``email_id`` in the request body as the ID. The benefit is we don't have to parse the response, either the body or the Location header.

If we want CTMS to pick the ID, we can omit the ``email_id`` in the body, CTMS will generate one. We'll need to decide if ``CTMS.add()`` returns just the ``email_id``, the full JSON, or both.

basket/news/backends/ctms.py
```diff
@@ -101,6 +106,170 @@ def from_vendor(contact):
...
+    "fpn_country": partial(truncate_string, 100),  # SFDC was 120
+    "fpn_platform": partial(truncate_string, 100),  # SFDC was 120
+}
+
+
+def to_vendor(data):
+    """
+    Transform basket key-value data and convert to CTMS nested data
+
+    Differences from SFDC.to_vendor:
```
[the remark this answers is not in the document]
jwhitlock: You can tell how often I've been burned by bad comments in my previous code.

PR #688, opened 2021-04-14
Generate email IDs in CTMS, use safer calls

### LINKED ISSUE #686: CTMS should generate the email ID for new contacts
CTMS email IDs are v4 UUIDs, so they could be generated from Basket or from CTMS. However, if this changes in the future, Basket would need to update as well. To avoid problems, let's let CTMS generate the email_id and Basket can read it.

jwhitlock:
Fix #686:

* Instead of generating the email ID in Basket, let CTMS generate it, and pass it to SFDC when creating the contact.
* Surround more CTMS calls with exception handlers.
* Re-arrange ``sfdc_add_update`` to also create the CTMS record before it tries to create the SFDC record.

### DIFF
--- a/basket/news/tasks.py
+++ b/basket/news/tasks.py
@@ -564,14 +564,13 @@ def upsert_contact(api_call_type, data, user_data):
             sfdc_add_update.delay(update_data)
         else:
             ctms_data = update_data.copy()
-            email_id = ctms_data["email_id"] = generate_token()
             try:
-                ctms.add(ctms_data)
+                ctms_contact = ctms.add(ctms_data)
             except Exception:
                 sentry_sdk.capture_exception()
             else:
                 # Successfully added to CTMS, send email_id to SFDC
-                update_data["email_id"] = email_id
+                update_data["email_id"] = ctms_contact["email"]["email_id"]
 
             # don't catch exceptions here. SalesforceError subclasses will retry.
             sfdc.add(update_data)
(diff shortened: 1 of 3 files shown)


basket/news/tasks.py
```diff
@@ -642,17 +650,23 @@ def sfdc_add_update(update_data, user_data=None):
...
+                    except Exception:
+                        sentry_sdk.capture_exception()
             else:
                 # still no user, try the add one more time
+                try:
+                    ctms_contact = ctms.add(update_data)
+                except Exception:
+                    sentry_sdk.capture_exception()
+                else:
+                    update_data["email_id"] = ctms_contact["email_id"]["email_id"]
```
pmac, line 668: shouldn't this be?
```suggestion
                    update_data["email_id"] = ctms_contact["email"]["email_id"]
```

jwhitlock:
I've got a couple changes I'm considering:

1. Move the `try`/`except` blocks into `CTMS`. If the CTMS call raises an error, then send to sentry and return `None`. This will reduce the redundant `try`/`except` blocks as I convert more code to talk to CTMS. However, we'll probably need to raise and handle exceptions later, as CTMS becomes the primary. when we need to deal with ID collisions and other issues.
2. Run `from_vendor` on the return from `ctms.add` and `ctms.update`. This may waste some CPU cycles, but keeps the basket code working in the basket format, rather than some of the CTMS format leaking into the other code.

With these two changes, this code:

```python
try:
  ctms_contact = ctms.add(update_data)
except Exception:
  sentry_sdk.capture_exception()
else:
  update_data["email_id"] = ctms_contact["email"]["email_id"]
```

would become:

```python
ctms_contact = ctms.add(update_data)
if ctms_contact:
  update_data["email_id"] = ctms_contact["email_id"]
```

or, since we're on Python 3.8, even:
```python
if (ctms_contact := ctms.add(update_data)):
  update_data["email_id"] = ctms_contact["email_id"]
```

I'm open to other ideas.

pmac:
> Move the `try`/`except` blocks into `CTMS`.

By this you mean move the blocks into the CTMS backend code? i.e. move them into `ctms.add`? That sounds good. 

The walrus operator looks fun. Haven't used it yet, but you're right we could.

jwhitlock:
> By this you mean move the blocks into the CTMS backend code? i.e. move them into ctms.add? That sounds good.

Yes, that's what I meant. I'll do that one first, see what that looks like.

jwhitlock:
`flake8` doesn't like the walrus operator:

```
"pyflakes" failed during execution due to "'FlakesChecker' object has no attribute 'NAMEDEXPR'"
```

I've moved the `try`/`except` blocks but didn't pass the output through `to_vendor`. I'm less sure about that one.

pmac:
> `flake8` doesn't like the walrus operator:

That's fine. We could probably upgrade that or do something to allow it, but I'm sure some people still use 3.7 in places (I'm pretty sure one such person is me). So staying with 3.7 compat is probably good anyway.

pmac:
This looks good to me. Are you done and ready to merge?

jwhitlock:
Yes, ready to merge!

Issue #684, opened 2021-04-14
Update additional basket methods to talk to CTMS


jwhitlock:
Update additional basket methods to talk to CTMS:

* [x] [confirm_user](https://github.com/mozmeao/basket/blob/922d6b50083ed470047d43ac5684883f180142c8/basket/news/tasks.py#L656) - Called when a user confirms a newsletter subscription. Add a parallel update to CTMS.
* [ ] [fxa_delete](https://github.com/mozmeao/basket/blob/57b2b53799b4061055c7873f12b95cfcc03022df/basket/news/tasks.py#L324) - Called when an FxA account is deleted. Add a parallel update to CTMS.
* [ ] [fxa_email_changed](https://github.com/mozmeao/basket/blob/57b2b53799b4061055c7873f12b95cfcc03022df/basket/news/tasks.py#L279) - Called when a user changes their FxA account emailed. Add a parallel update to CTMS.
* [ ] [process_donation](https://github.com/mozmeao/basket/blob/57b2b53799b4061055c7873f12b95cfcc03022df/basket/news/tasks.py#L913) - Called when processing a donation to the Mozilla Foundation. Add a parallel update to CTMS.
* [ ] [process_petition_signature](https://github.com/mozmeao/basket/blob/57b2b53799b4061055c7873f12b95cfcc03022df/basket/news/tasks.py#L1098) - Called when processing a petition signature. Add a parallel update to CTMS.
* [x] [record_common_voice_update](https://github.com/mozmeao/basket/blob/bc8651db0cf67a4eeea26a660826beafad02a94d/basket/news/tasks.py#L731) - Called when processing a Common Voice update, which may include subscribing to the newsletter. Add a parallel update to CTMS.
* [x] [update_custom_unsub](https://github.com/mozmeao/basket/blob/922d6b50083ed470047d43ac5684883f180142c8/basket/news/tasks.py#L702) - Called when a user gives a reason for unsubscribing from all newsletters. Add a call to populate unknown email_ids from CTMS, or to update by basket token.
* [x] [update_user_meta](https://github.com/mozmeao/basket/blob/bc8651db0cf67a4eeea26a660826beafad02a94d/basket/news/tasks.py#L471) - Called from ``/news/user-meta/{token}/``. Add a call to populate unknown email_ids from CTMS, or to update by basket token.
[...]

PR #689, opened 2021-04-21
Adjust CTMS interface, integrate more CTMS calls

jwhitlock:
This PR makes several changes, which could be reviewed by commit:

* Adjust and update the CTMS interface:
  - Add ``CTMSError`` as the base exception for CTMS exceptions, as well as further derived exceptions. This work is not complete, and some exceptions are still raised directly from ``requests``. The goal is to start adding some exception handling into the basket code.
  - ``ctms.update`` now does nothing if ``email_id`` isn't set. This simplifies the calling code to look more like what it should after SFDC is retired
  - New ``ctms.update_by_alt_id`` method. This combines fetching a record by an ID other than ``email_id``, like the basket token, and then running a partial update. This may become a single API call  if CTMS grows another endpoint.
* Add basic tests for untested tasks ``update_custom_unsub`` and ``update_user_meta``
* Update tasks to also call CTMS:
  - ``confirm_user`` - Also try to update CTMS record
  - ``update_custom_unsub`` - Also try to update CTMS record
  - ``update_user_meta`` - Also try to update CTMS record
  - ``record_common_voice_update`` - Also create or (try to) update CTMS record
* Allow ``pytest`` as well as ``py.test`` to adjust Django settings for tests. 

This makes progress on issue #684, but leaves several tasks still to convert (called by ``process_donations_queue``, ``process_fxa_queue``, and ``amo_sync_addon``)

### DIFF
--- a/basket/news/backends/ctms.py
+++ b/basket/news/backends/ctms.py
@@ -23,6 +23,11 @@ from basket.news.newsletters import (
 
 time_request = get_timer_decorator("news.backends.ctms")
 
+
+class CTMSError(Exception):
+    """Base class for exceptions raised by CTMS functions."""
+
+
 # Map CTMS group / name pairs to basket flat format names
 # Missing basket names from pre-2021 SFDC integration:
 #  record_type - Identify MoFo donors with an associated opportunity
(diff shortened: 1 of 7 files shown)


pmac (approved):
🤘

PR #700, opened 2021-04-30
Add .is_mofo to newsletters to identify users

jwhitlock:
Add a column ``is_mofo`` (default ``False``) to newsletters, to signal that a newsletter is MoFo-related. When a user subscribes to one of these newsletters, set the ``mofo_relevant`` flag if it is not yet set.


pmac (approved):
🏛️

Issue #697, opened 2021-04-29
De-prioritize SFDC ID when getting user data from CTMS


jwhitlock:
When creating a new contact, it is first created in CTMS before being created in SFDC. This means that CTMS doesn't get the SFDC ID until and unless it is added by a later update. In the current code, however, the SFDC ID is considered the third-best option, after basket token and email, and before FxA ID and AMO ID. CTMS should try looking up by SFDC ID last, and should possibly try by multiple IDs rather than just the top-priority ID.

jwhitlock:
Fixed with PR #702

PR #702, opened 2021-05-07
CTMS.get: Try SFDC ID last, and try multiple IDs

jwhitlock:
The SFDC ID is being tried before the AMO and FxA IDs, which may be resulting in more CTMS misses than anticipated. In the BQ import, some records have multiple SFDC IDs, so it is no longer treated as unique.

When Basket calls ``CTMS.get`` with multiple alternate IDs (not CTMS's primary `email_id`), call CTMS for each ID value, and return the first result with a single contact.

This should fix issue #697

### DIFF
--- a/basket/news/backends/ctms.py
+++ b/basket/news/backends/ctms.py
@@ -656,19 +656,20 @@ class CTMS:
         if email_id:
             contact = self.interface.get_by_email_id(email_id)
         else:
+            alt_ids = []
             if token:
-                id_name, id_value = "basket_token", token
-            elif email:
-                id_name, id_value = "primary_email", email
-            elif fxa_id:
-                id_name, id_value = "fxa_id", fxa_id
-            elif mofo_email_id:
-                id_name, id_value = "mofo_email_id", mofo_email_id
-            elif amo_id:
-                id_name, id_value = "amo_user_id", amo_id
-            elif sfdc_id:
-                id_name, id_value = "sfdc_id", sfdc_id
-            else:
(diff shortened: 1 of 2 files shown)


basket/news/backends/ctms.py
```diff
@@ -682,15 +682,22 @@ def get(
...
-                contact = contacts[0]
-            else:
-                raise CTMSMultipleContactsError(id_name, id_value, contacts)
+            # Try alternate IDs in order
+            first_run = True
+            first_contacts = None
+            contact = None
+            for params in alt_ids:
+                contacts = self.interface.get_by_alternate_id(**params)
+                if first_run:
```
pmac, line 691: why would it only be the first one that could return multiples?

basket/news/backends/ctms.py
```diff
@@ -682,15 +682,22 @@ def get(
...
+                    first_contacts = contacts
+                    first_run = False
+                if len(contacts) == 1:
+                    contact = contacts[0]
```
pmac, line 695: shouldn't we have a `break` here once we find a good contact?

basket/news/backends/ctms.py
jwhitlock: I'm not sure there is a best return if multiple IDs checks fail, because each returned no contacts or multiple contacts. So, it returns the result of the first one - a ``CTMSMultipleContactsError`` if the first one returned multiple, or ``None`` if it returned no contact list. This matches the previous behavior, when it would check one of the IDs and then give up.

PR #705, opened 2021-05-10
CTMS: Set the source URLs for dict newsletters

jwhitlock:
When newsletter subscriptions are sent as a dictionary, also set the source URL for new subscriptions. Previously, the source URL was only set when the subscriptions were a list.


jwhitlock:
This hopefully addresses this bug from https://github.com/mozilla-it/ctms-api/issues/184:

> CTMS is not populating the "source" field on the "newsletter" field. Basket should be providing this.

PR #708, opened 2021-05-17
Updating call to CTMS when unsubscribe all is called. Added email_schema body to request as well.

bsieber-mozilla:
@pmac, @jwhitlock , @cvalaas

This PR modifies the conditional behavior for OPTOPT. 
If optopt is present this should produce an unsubscribe event.

Since the PatchSchema is expecting email changes as well, the request could contain the "has_opted_out_of_email" to be updated by CTMS through the ORM into the DB

### DIFF
--- a/basket/news/backends/ctms.py
+++ b/basket/news/backends/ctms.py
@@ -275,10 +275,10 @@ def to_vendor(data, existing_data=None):
         if isinstance(newsletters, dict):
             # Detect unsubscribe all
             optout = data.get("optout", False) or False
-            secondary_check = not any(newsletters.values()) and set(valid_slugs) == set(newsletters.keys())
-            if (
-                optout or secondary_check
-            ):
+            secondary_check = not any(newsletters.values()) and set(valid_slugs) == set(
+                newsletters.keys(),
+            )
+            if optout or secondary_check:
                 # When unsubscribe all is requested, let CTMS unsubscribe from all
                 output = "UNSUBSCRIBE"
 
@@ -287,9 +287,7 @@ def to_vendor(data, existing_data=None):
                     ctms_data["email"] = {}
                 elif ctms_data["email"] is None:
                     ctms_data["email"] = {}
-
                 ctms_data["email"]["has_opted_out_of_email"] = True
-
             else:
                 # Dictionary of slugs to sub/unsub flags
                 for slug, subscribed in newsletters.items():
@@ -383,10 +381,10 @@ class CTMSSession:
             token_updater=self.save_token,
         )
         session.register_compliance_hook(
-            "access_token_response", CTMSSession.check_2xx_response
+            "access_token_response", CTMSSession.check_2xx_response,
         )
         session.register_compliance_hook(
-            "refresh_token_response", CTMSSession.check_2xx_response
+            "refresh_token_response", CTMSSession.check_2xx_response,
         )
         if not session.authorized:
             session = self._authorize_session(session)
@@ -602,7 +600,7 @@ class CTMSMultipleContactsError(CTMSError):
     def __str__(self):
         try:
             email_ids = repr(
-                [contact["email"]["email_id"] for contact in self.contacts]
+                [contact["email"]["email_id"] for contact in self.contacts],
             )
         except Exception:
             email_ids = "(unable to extract email_ids)"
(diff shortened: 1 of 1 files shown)


basket/news/backends/ctms.py
```diff
@@ -375,10 +377,10 @@ def _session(self):
             token_updater=self.save_token,
         )
         session.register_compliance_hook(
-            "access_token_response", CTMSSession.check_2xx_response
+            "access_token_response", CTMSSession.check_2xx_response,
```
bsieber-mozilla, line 380: _precommit_

basket/news/backends/ctms.py
```diff
@@ -375,10 +377,10 @@ def _session(self):
...
         )
         session.register_compliance_hook(
-            "refresh_token_response", CTMSSession.check_2xx_response
+            "refresh_token_response", CTMSSession.check_2xx_response,
```
bsieber-mozilla, line 383: _precommit_

basket/news/backends/ctms.py
```diff
@@ -594,7 +596,7 @@ def __repr__(self):
     def __str__(self):
         try:
             email_ids = repr(
-                [contact["email"]["email_id"] for contact in self.contacts]
+                [contact["email"]["email_id"] for contact in self.contacts],
```
bsieber-mozilla, line 599: _precommit_

basket/news/backends/ctms.py
```diff
@@ -275,13 +275,15 @@ def to_vendor(data, existing_data=None):
...
-            ):
+            if optout:
                 # When unsubscribe all is requested, let CTMS unsubscribe from all
                 output = "UNSUBSCRIBE"
+                # Set has_opted_out_of_email for email
+                if "email" not in ctms_data.keys():
+                    ctms_data["email"] = {}
+                elif ctms_data["email"] is None:
+                    ctms_data["email"] = {}
+                ctms_data["email"]["has_opted_out_of_email"] = True
```
jwhitlock, line 286: I'm surprised this is needed, it should have been set up at line 255. But, since there are no changes to tests, Is suspect it does nothing.

bsieber-mozilla:
The second "Triggering pre-commit" commit should be "Fixing tests. Removing second conditional check to separate ubsubscribe_all from all newsletters unsubscribe each."

Hopefully the the doubled message, doesn't cause a problem.

basket/news/backends/ctms.py
```diff
@@ -275,13 +275,15 @@ def to_vendor(data, existing_data=None):
         if isinstance(newsletters, dict):
             # Detect unsubscribe all
             optout = data.get("optout", False) or False
-            if (
-                optout
-                and (not any(newsletters.values()))
-                and (set(valid_slugs) == set(newsletters.keys()))
-            ):
+            if optout:
```
jwhitlock, line 278: Does this prevent someone from re-subscribing after they've opted out once? I think it is OK since `optout` would only be sent when unsubscribing, but it is worth integration testing since we don't know what basket sends.

basket/news/backends/ctms.py
jwhitlock: sloppy. If precommit is needlessly changing code, then it should be disabled.

jwhitlock (approved):
I am surprised I have the ability to approve PRs!

Thanks @bsieber-mozilla. I'm not confident that this will hurt or help the issue, but it also doesn't break tests, so it probably won't cause more damage. Let's give it a try.

basket/news/backends/ctms.py
bsieber-mozilla: I believe you are correct. After this PR, I went back to re-read some of the code and tests. I quickly drafted a better PR I believe here///: https://github.com/mozmeao/basket/pull/709

basket/news/backends/ctms.py
bsieber-mozilla: I agree we'll need to integration test this. As long as that value isn't repeatedly sent, after a re-subscribe event the newsletters should have the ability to be re-subscribed.

bsieber-mozilla:
_Hold on merge for a bit longer._

bsieber-mozilla:
Opting for: #709 
Will close this shortly.

basket/news/backends/ctms.py
bsieber-mozilla: Moved this earlier in the new pr.

basket/news/backends/ctms.py
bsieber-mozilla: Removed this logic in the new pr

PR #709, opened 2021-05-17
Updating the Request to CTMS when Empty Newsletter w/ Optout

bsieber-mozilla:


bsieber-mozilla:
#Adding test scenario
In this commit I added a scenario that currently is not working in the system.

#Updating Conditional
In this commit I've updated the conditional to capture this additional scenario.

bsieber-mozilla:
#Updating value at time of conditional
To ensure that the unsubscribe was being provided, when not seen in new test (valid?) scenario.

#Moving optout check to before newsletter check 
Moved the conditional to before checking for existence of newsletters.

jwhitlock (approved):
Thanks @bsieber-mozilla, I appreciate the targeted test

bsieber-mozilla:
👍 wfm; the added test might not be valid if the request to optout always
came with a newsletter object.

Issue #696, opened 2021-04-29
Implement CTMS-only mode


jwhitlock:
``CTMS_ENABLED=True`` is implemented in dual-write mode, where SFDC is the primary source, and CTMS gets a copy of most updates. In the final mode, CTMS will be the only source of data, SFDC will not be consulted.  This could be implemented with a ``SFDC_ENABLED`` flag, default on, which can be used to turn off the SFDC integration. This could be tested in `basket-dev`, and used to make the transition in May.

There may be another transition phase, such as ``CTMS_FIRST=True``, where both systems are still consulted, but CTMS is the primary source of user data.

jwhitlock:
Implemented with ``SFDC_ENABLED=False``, and deployed to dev, stage, and prod.

PR #710, opened 2021-05-18
Implement CTMS-Only mode with `SFDC_ENABLED=False`

jwhitlock:
For issue #696, implement CTMS-Only mode by adding a new setting `SFDC_ENABLED`, which defaults to `True` if SFDC_USERNAME is set. When it is `False`, several things change:

* ``backends.sfdc.sfdc`` raises a new error ``SFDCDisabled`` if ``sfdc.get`` is called or a low-level SFDC type class is allocated (like ``sfdc.contact``).
* ``backends.ctms.ctms`` raises a new error ``CTMSNotConfigured`` if methods are called and the CTMS API key is not configured.
* ``utils.get_user_data`` catches ``SFDCDisabled`` and uses CTMS as the primary user data source. Several exceptions are re-raised as ``NewsletterException`` instead of being captured and ignored. The ignored-fields logic now runs against the CTMS user data as well.
* Several tasks change implicitly because user data comes from CTMS, and ``sfdc.add`` and ``sfdc.update`` do nothing.
* ``task.update_user_meta`` raises an exception if a matching CTMS record is not found
* ``process_donation_event`` sends an email to ``DONATE_NOTIFY_EMAIL``
* ``process_donation`` skips adding opportunities to SFDC
* ``process_petition_signature`` skips adding campaign members to SFDC
* ``amo_sync_addon`` does not process any data

### DIFF
--- a/basket/news/backends/ctms.py
+++ b/basket/news/backends/ctms.py
@@ -618,11 +618,19 @@ class CTMSNotFoundByAltIDError(CTMSError):
         return f"No contacts returned for {self.id_name}={self.id_value!r}"
 
 
+class CTMSNotConfigured(CTMSError):
+    """CTMS is not configured."""
+
+    def __str__(self):
+        return "CTMS is not configured" ""
+
+
 class CTMS:
     """Basket interface to CTMS"""
 
-    def __init__(self, interface):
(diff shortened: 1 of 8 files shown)


pmac (approved):
:shipit:

PR #712, opened 2021-05-20
ctms.to_vendor: Log in sentry rather than raise exception on unknown keys

jwhitlock:
Previously, `CTMSUnknownKeyError` was raised when the Basket update data had an unrecognized key. Instead, the code now collects these and uses ``sentry_sdk.capture_message`` to log it in Sentry but continue processing.

Two valid keys (`api-key` and `trigger_welcome`) that caused issues in production are now in the `DISCARD_BASKET_NAMES` set, and will be silently discarded.

### DIFF
--- a/basket/news/backends/ctms.py
+++ b/basket/news/backends/ctms.py
@@ -193,19 +193,6 @@ TO_VENDOR_PROCESSORS = {
 }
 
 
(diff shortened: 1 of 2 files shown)


pmac (approved):
🌌

Issue #718, opened 2021-06-04
Retry CTMS updates for network issues


jwhitlock:
Use the existing ``et_task`` framework to catch exceptions from CTMS API calls, and reschedule tasks due to temporary network issues.

This will require raising exceptions from ``basket.news.backends.ctms.ctms``, rather than reporting to Sentry and returning ``None``. It will also require distinguishing between retryable exceptions like ``503 Service Gateway`` errors, and non-retryable exceptions like ``404 Not Found``, ``409 Conflict``, and ``422 Unprocessable Entity``.

Issue #719, opened 2021-06-04
Remove Salesforce / SFDC interface code


jwhitlock:
Remove the Salesforce interface code from Basket:

* Remove paths for ``settings.SFDC_ENABLED``, assume not enabled
* Remove calls to ``basket.news.backends.sfdc.sfdc``, retaining the ones to ``basket.news.backends.ctms.ctms``
* Remove SFDC-only code paths, such as donation payment updates, or raise an exception if called
* Convert ``fxa_delete`` to non-generic CTMS code

When there is no more code that calls it:
* Drop the SFDC backend
* Drop Salesforce settings
* Rename tasks with `sfdc` in title

Also tracked as https://jira.mozilla.com/browse/OSS-435

robhudson:
This work has been completed.

PR #720, opened 2021-06-04
Raise exceptions from CTMS backend, to enable retries in et_task

### LINKED ISSUE #713: Include error JSON in CTMS errors
CTMS reports failed statuses with ``raise_for_status``, like:

https://github.com/mozmeao/basket/blob/ce9439490a6554ee85ed55f21cdec882c4a3ec67/basket/news/backends/ctms.py#L569-L571

In Sentry, this reports:
* The status code (such as `422`)
* The reason (such as `"Unprocessable Entity"`)
* A summary message (such as `"422 Client Error: Unprocessable Entity for url: https://ctms.stage.mozilla-ess.mozit.cloud/ctms/{email_id}"`

For a 422 and other errors, the JSON body of the response also has information, which is useful for diagnosing the error, and should be included in the context sent to Sentry

jwhitlock:
This PR fixes issue #718.

Previously, exceptions would be captured in the ``CTMS`` class, reported to Sentry, and then ``None`` would be returned to the caller. Instead, let the exceptions be returned to the caller.

``requests`` raises exceptions derived from a shared base exception class ``RequestException``, which is derived from ``IOError``. This means that low-level exceptions such as timeouts and DNS lookup failures will be retried by ``et_task`` with no code changes. 

It also means HTTP Errors such as ``400 Bad Request`` and ``503 Service Unavailable`` will be retried. Some HTTP errors represent issues that will not be fixed by a Retry, so these are now raised as new exceptions, which will not be retried by ``et_task``:

* ``CTMSNotFoundByEmailIDError`` - Replaces a 404 when the ``{email_id}`` is part of the URL
* ``CTMSUniqueIDConflictError`` - Replaces a 409 error, signifying a primary ID conflict
* ``CTMSValidationError`` - Replaces a 422 error, signifying a schema validation error. This also fixes #713.

Most utility and task code did not need changes to propagate these exceptions. The two changes are:

* ``get_user_data`` - Remove the 404 exception, since the ``/ctms`` endpoint doesn't return a 404 on no match.Instead, handle the ``CTMSNotFoundByAltIDError`` exception by returning None, similar to the SFDC code.
* ``sfdc_add_update`` - Remove all the SFDC interface code (getting started on issue #719), and handle switching from add to and update when ``CTMSUniqueIDConflictError`` is raised.

### DIFF
--- a/basket/news/backends/ctms.py
+++ b/basket/news/backends/ctms.py
@@ -800,11 +800,7 @@ class CTMS:
                 raise CTMSNotConfigured()
             else:
                 return None
-        try:
-            return self.interface.post_to_create(to_vendor(data))
-        except Exception:
-            sentry_sdk.capture_exception()
-            return None
+        return self.interface.post_to_create(to_vendor(data))
 
     def update(self, existing_data, update_data):
         """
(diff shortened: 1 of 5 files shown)


pmac (approved):
🍰

PR #781, opened 2022-01-17
Disable logging of DisallowedHost to reduce log noise

stevejalim:
Django will still respond with HTTP 400 to requests with unexpected hostnames, but we won't get the log noise.

Also had to `black` the relevant file to keep `pre-commit` happy.


pmac (approved):
💯

Issue #784, opened 2022-01-24
Split 


stevejalim:
Once #782 is done, we will have separate prod and dev requirements files. It would be good to leverage these to make our prod build smaller and with no unnecesary dependencies.

To acheive this, we will need to refactor our Docker setup for Basket, to generate a Dev/Test image separate from the main/Prod image and to use these as appropriate

### Success criteria
- [ ] Current container/image workflow is understood
- [ ] Workflow is updated to generate dev-and-test-specific and prod-specific builds
- [ ] Prod build still gets shipped to basket-production and basket-staging
- [ ] Dev/Test build gets shipped to basket-dev
- [ ] CI is still happy

PR #786, opened 2022-01-24
Move to pip-compile-multi

stevejalim:
This changeset swaps Basket form pip-tools to using pip-compile-multi as our chosen dependency helper, plus retains and evolves a Makefile-driven approach to using pip-compile-multi.

Notes: 

* UPDATED: this continues to compile requirements inside the Docker container

* for now, while we generate separate `requirements/prod.txt` and `requirements/dev.txt` files, we continue to use the superset of dependencies (requirements/dev.txt) for building the single Docker image. Splitting Docker images is deferred to #784 to keep this change cleaner.

* one dependency - the FuelSDK for Salesforce - had to be removed as it has a very stale dependency and patching the supply chain isn't worth it as FuelSDK isn't used in the Basket codebase any more, it appears.

### To test
* Run `make compile-requirements` 
* Run `make compile-requirements` again and confirm that the deps are rebuilt, but have no changes
* Run `make clean build test` to ensure that full images still build
* Check CI is happy


Makefile
```diff
@@ -59,11 +63,14 @@ test: .make.docker.pull
...
+compile-requirements: .make.pip-compile-multi.check-installed
+	./bin/compile-requirements.sh
+
+upgrade-requirements: .make.pip-compile-multi.check-installed
+	./bin/compile-requirements.sh --upgrade
 
-upgrade-requirements: .make.docker.pull
-	${DC} run --rm upgrade-requirements
+check-requirements:
+	pip list -o
```
robhudson, line 73: This has to run in the container.

bin/compile-requirements.sh
```diff
@@ -4,7 +4,7 @@ set -exo pipefail
 
 export CUSTOM_COMPILE_COMMAND="make compile-requirements"
 if [[ "$1" == "--upgrade" ]]; then
-    pip-compile --generate-hashes --reuse-hashes --upgrade --upgrade-package 'django<2.3' requirements.in
+    pip-compile-multi --generate-hashes prod --generate-hashes dev --upgrade
 else
-    pip-compile --generate-hashes --reuse-hashes requirements.in
+    pip-compile-multi --generate-hashes prod --generate-hashes dev
```
robhudson, line 9: If my host Python version is different that Docker's, will this still produce the correct results? I know some sub-dependencies get installed depending on Python version.

docs/install.rst
```diff
@@ -79,3 +79,18 @@ And if you need to debug a running container, you can open another terminal to y
...
 
+
+Maintaining Python requirements
+-------------------------------
+
+.. code-block:: bash
+
+    $ # If you've added a new dependency
+    $ make compile-requirements
+    $ # or if you wantt upgrade all dependencies
```
robhudson, line 90: "wantt"

robhudson (approved):
Just some questions, otherwise looks good. r+wc

robhudson:
Hmm, this failed for me b/c I don't have mysql on my host machine:
`    OSError: mysql_config not found`

Makefile
stevejalim: - [ ] move to run in the container

bin/compile-requirements.sh
stevejalim: Yeah, this is the thing between running this all in the container or living with a mix of local and containerised... I didn't like the idea of having pip-tools itself in the dependency manifests, in part because a stale pip-tools in hte project created a circular dependency problem where I couldn't update pip tools to update other deps. So, I moved it out, but can move it back in to run in the container, sure

In theory, if you get an inviable dep mix because you updated them on, say, 3.9 on your host, the 3.7 container should fail to install. However, this isn't ideal, I know, so let's stabilise it

- [ ] @stevejalim move to run in the container

stevejalim:
@robhudson @pmac One additional point is that `make docs` fails for me both on this branch and on `master`. Would one of you mind trying it to see if you can reproduce?
```Error response from daemon: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "make": executable file not found in $PATH: unknown```

File: bin/compile-requirements.sh [hunk 4]
```diff
@@ -3,8 +3,15 @@
...
 
 export CUSTOM_COMPILE_COMMAND="make compile-requirements"
+
+# We need this installed, but we don't want it to live in the main requirements
+# We will need to periodically review this pinning
+pip install --upgrade pip-tools==6.4.0  # needs at least this version to build
+pip install pip-compile-multi
+
+
 if [[ "$1" == "--upgrade" ]]; then
```
robhudson, line 13: Similar note about `--upgrade` as on the bedrock PR

robhudson (approved):
Looks good to me.

requirements/prod.txt
```diff
@@ -1,130 +1,152 @@
...
+    --hash=sha256:e55b384612d93be96506932a786bbcde5a2db7a9e6a4bb4bffe8b733f5b9036b \
+    --hash=sha256:e81b76cace8eda1fca50e345242ba977f9be6ae3945af8d46326d776b4cf78d1 \
+    --hash=sha256:e8236383a20872c0cdf5a62b554b27538db7fa1bbec52429d8d106effbaeca08 \
+    --hash=sha256:f04e857b59d9d1ccc39ce2da1021d196e47234873820cbeaad210724b1ee28ac \
+    --hash=sha256:fadbfe37f74051d024037f223b8e001611eac868b5c5b06144ef4d8b799862f2
+    # via
+    #   -r requirements/prod.in
+    #   pytz-deprecation-shim
+    #   tzlocal
+billiard==3.6.4.0 \
```
pmac, line 41: We should just double check that this version is compatible with our version of Celery.

pmac (approved):
Looks great! 🥇

PR #800, opened 2022-03-04
Adding Relay Waitlist to CTMS API Call

bsieber-mozilla:
Looking to onboard the Relay team's waitlist into Acoustic and CTMS.

### DIFF
--- a/basket/news/backends/ctms.py
+++ b/basket/news/backends/ctms.py
@@ -83,6 +83,7 @@ CTMS_TO_BASKET_NAMES = {
         "mofo_relevant": "mofo_relevant",
     },
     "vpn_waitlist": {"geo": "fpn_country", "platform": "fpn_platform"},
+    "relay_waitlist": {"geo": "relay_country"},
 }
 
 
--- a/basket/news/tests/test_ctms.py
+++ b/basket/news/tests/test_ctms.py
@@ -84,6 +84,9 @@ SAMPLE_CTMS_RESPONSE = json.loads(
     "geo": "fr",
     "platform": "ios,mac"
   },
+  "relay_waitlist": {
+    "geo": "fr",
+  },
   "status": "ok"
 }
 """
@@ -104,6 +107,7 @@ SAMPLE_BASKET_FORMAT = {
     "format": "H",
     "fpn_country": "fr",
     "fpn_platform": "ios,mac",
+    "relay_country": "fr",
     "fxa_create_date": "2021-01-29T18:43:49.082375+00:00",
     "fxa_deleted": False,
     "fxa_id": "6eb6ed6ac3b64259968aa490c6c0b9df",


basket/news/backends/ctms.py
```diff
@@ -83,6 +83,7 @@ class CTMSError(Exception):
         "mofo_relevant": "mofo_relevant",
     },
     "vpn_waitlist": {"geo": "fpn_country", "platform": "fpn_platform"},
+    "relay_waitlist": {"geo": "relay_country"},
```
leplatrem, line 86: Add this `relay_country` to the tests so that we can test we will receive it as `{"relay_waitlist": {"geo": "XX"}}`
[    "fpn_country": "fr",](https://github.com/mozmeao/basket/blob/8b6f57379875f9d8c94428e3cd6be2af9f3b9152/basket/news/tests/test_ctms.py#L105)

bsieber-mozilla:
We are hoping to add an additional waitlist for the relay team; they plan on sending their form data to basket with the request param `relay_country`. 

What is the process for PR review + deploy look like? @duallain @pmac

leplatrem (approved):

stevejalim (approved):
LGTM on the surface!

pmac:
Are y'all also going to do the work to have basket accept the `relay_country` param for subscribe submissions, or do we need to add that to our backlog?

pmac:
Do you know what API endpoint relay is hitting on basket? `/news/subscribe/`?

bsieber-mozilla:
> Do you know what API endpoint relay is hitting on basket? `/news/subscribe/`?

I believe that is the endpoint that they are planning on hitting; we looked into the VPN logic a bit which appeared to use [subscribe api](https://github.com/mozilla/bedrock/blob/857129a9089bc2927e797a776880ab3b9f6d9f9a/bedrock/products/views.py#L84-L93) through bedrock.

Relay team might need an API key as well to interact with Basket?  _(...unless they also choose to go through Bedrock)_

pmac:
They only need an API key to use with basket if we set the waitlist as "private", which means that an API key is required to subscribe. But as you pointed out bedrock is supplying that key for the VPN form on bedrock. That form is public though, so anyone could post to it and sign up. I don't see much point in setting the "private" flag on the list so Relay can just hit basket directly from their frontend. If they're doing a similar subscribe thing then I can ensure those forms accept the new value. 

I'm also curious why we don't just use the country field we already have for this? It seems that a users country is more of a property of a contact rather than a subscription.

bsieber-mozilla:
@pmac thanks for your response; let me check with the others and get back to you on this soon

leplatrem:
I also understood that the API key allows to lookup an existing subscription by email. 

I saw that in Bedrock, where it does a `get_user(email, api_key)` on Basket before updating its subscriptions.

But maybe Relay does not have to go that far, and would be fine with just a `POST /news/subscribe`

pmac:
> But maybe Relay does not have to go that far, and would be fine with just a `POST /news/subscribe`

For subscribing to anything (newsletter, waitlist, etc.) just POSTing to subscribe is the right thing. It's not necessary to check if they're subscribed beforehand. The subscription call is idempotent.

pmac:
You could be right. I just though initially that there'd be more. I'll make sure and get back to you.

pmac:
Yes, I think you are right. My apologies for the confusion. We should test and make sure as well of course.

PR #801, opened 2022-03-08
Upgrade basket to Python 3.9.10

### LINKED ISSUE #788: Upgrade to Python 3.9
As part of updating all MEAO's codebases, we should move Basket from Python 3.8 to 3.9 (or even 3.10 if viable)

### Success critera
- [ ] all unit tests passing on 3.9
- [ ] manual test-drive OK on 3.9
- [ ] no new 3.9-related regressions in Sentry first 24 hours after release

stevejalim:
This changeset bumps basket from 3.8 to 3.9.10 and also updates its dependencies (where safe/stable) to be versions which definitely support 3.9+

Resolves #788 

### Success criteria

- [ ] All tests passing
- [ ] No regressions on dev, stage or prod


stevejalim:
@pmac There's one controversial/debatable thing here, which is that migration is generated for the `LocaleField` that drops all the default choices from the frozen state of the model. I

I think this is because the `django-product-details` dependency (which is used to generate the default list of `LocaleField` options) hasn't had its data bootstrapped. 
 
If we _don't_ include this migration, _or_ we include one which is based on the very latest state of the locale data, checks/tests fail - so we need to be sure when we deploy that Django isn't going to find a 'warmed' cache of locale data in `product_details.languages` and decide that it needs _another_ migration before it'll boot. 

As such, we could try _now_ to move the choices into something that's more lazily evaulated, which should avoid that risk.

@pmac @robhudson what d'ya think? Have we had this before?

stevejalim:
@pmac  OK, this should be good to go now. Yes we still have had to add a migration, but it's because the locale set has increased in scope.

I know we talked about making the `choices` option super-lazily evaluated, but that might be a thing we leave till Django 3.2, when it'll be super simple

PR #802, opened 2022-03-08
Fix failing tests after CTMS addition

stevejalim:
Not sure why CI didn't catch this niggle when #800 was going through the early pipes

Also pre-commit hit the relevant test file with black formatting


basket/news/tests/test_ctms.py
```diff
@@ -85,11 +84,11 @@
     "platform": "ios,mac"
   },
   "relay_waitlist": {
-    "geo": "fr",
+    "geo": "fr"
```
pmac, line 87: _shakes fist at JSON_

PR #811, opened 2022-03-09
Upgrade to Django 3.2

robhudson:


basket/news/utils.py
```diff
@@ -135,7 +135,7 @@ def has_valid_api_key(request):
         or request.POST.get("api_key", None)
         or request.GET.get("api-key", None)
         or request.GET.get("api_key", None)
-        or request.META.get("HTTP_X_API_KEY", None)
```
stevejalim, line 138: Did this work with the old header name? Looks like it might not have?

basket/news/views.py
```diff
@@ -31,9 +30,7 @@
 from basket.news.models import (
     CommonVoiceUpdate,
     Interest,
-    LocaleStewards,
```
stevejalim, line 34: Are these two unimported models still in use elsewhere? If not, would be good too throw them in the woodchipper while we're at it?

stevejalim, line 34: Are these two unimported models still in use elsewhere? If not, would be good too throw them in the woodchipper while we're at it?

basket/settings.py
```diff
@@ -64,7 +64,7 @@ def path(*args):
     DATABASES["default"]["OPTIONS"] = {
         "init_command": "SET sql_mode='STRICT_TRANS_TABLES'",
     }
-
+DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
```
stevejalim, line 67: Nice one. I would have forgotten to add this!

stevejalim, line 67: Nice one. I would have forgotten to add this!

stevejalim (approved):
Looks good to me! ❤️

basket/news/views.py
robhudson: Yeah the are still in use, just no longer needed to be used with the synctool here. Good check though.

pmac (approved):
🏅

PR #813, opened 2022-03-14
Pre-filter Sentry event data before it goes over the network 

stevejalim:
This changeset uses [sentry-processor](https://pypi.org/project/sentry-processor/) to scrub any PII from payloads before it gets sent to Sentry. This is because we're moving from self-hosted Sentry to SaaS Sentry, and while I trust Sentry to scrub data as per our configuration and to remain secure, it's better to avoid any PII leaving Mozilla servers in the first place.

This changeset is not large, and the main work is some test coverage to confirm that `sentry-processor` is working as expected, given it has no public tests of its own. The tests are based on looking at the `sentry-processor` source and trying to get some smoke-test coverage in there, but they're not at all exhaustive - do say if you think this is a blocker.

Other things to consider when reviewing:

- [ ] `sentry-processor` is BSD licensed: does this play fine with the MPL2.0? I think it would do, but double checking
- [ ] Rather than zapping all sensitive fields, I've taken a swing at deciding what fields should be partially masked  - do say if you disagree. 
- [ ] Similarly, the partially masked fields only have four characters redacted. This is probably fine when it comes to redacting UUIDs etc, but am open to hiding the entire string.
- [ ] This changeset assumes that plugging in the processor to `sentry.init()` is enough to enable sanitisation on all things, so we'll have to check on Dev. (I did wonder about building `kent` into the overall Docker infra, but this seemed overkill for now)


basket/settings.py
```diff
@@ -341,11 +342,49 @@ def get_default_gateway_linux():
...
+
+def before_send(event, hint):
+    processor = DesensitizationProcessor(
+        with_default_keys=True,
+        sensitive_keys=SENSITIVE_FIELDS_TO_MASK_ENTIRELY,
+        partial_keys=SENSITIVE_FIELDS_TO_MASK_PARTIALLY,
+        mask_position=POSITION.LEFT,
+        off_set=3,
+    )
+    event = processor.process(event, hint)
```
pmac, line 378: style nit: any reason to not just `return processor.process(event, hint)` here?

pmac (approved):
Looks very nice! Thanks for all of the excellent work on tests!

basket/settings.py
stevejalim: Just a habit where the variable name helps hint at what's being returned, that's all.

PR #816, opened 2022-03-17
Tighten up PII scrubbing by no longer doing partial masking

stevejalim:
Instead, every sensitive string is fully redacted.

I've left the setup for the partial masking there, so that future selves can be reminded how to turn it back on, but we could just cut it if it's untidy


robhudson (approved):
lgmt 👍

PR #826, opened 2022-04-11
Upgrade Django following security release

stevejalim:


stevejalim:
@robhudson @pmac Just noticed that this branch doesn't appear to be running CI. Tests passed locally, but obvs it's nice to have CircleCI going too... I don't appear to have access to the Basket project in Circle, so maybe that's why I can't see it in GH?

PR #853, opened 2022-07-03
Dependency bumps, prompted by dependabot

stevejalim:
Note that `pytest-pythonpath` became unnecessary with pytest 7.0, but also it looks like we weren't using it anyway.

Tests are passing and changelogs for bumped versions mention no breaking changes.

Resolves #856
Resolves #855
Resolves #854 
Resolves #810
Resolves #808 
Resolves #807 
Resolves #805


robhudson (approved):
Looks good to me

PR #898, opened 2022-08-22
Remove SFDC

robhudson:
There is still more than can be done here but this is an interim stage that removes most and tests pass. I'll review deeper for more to remove or refactor.


basket/news/tasks.py
```diff
@@ -898,55 +846,31 @@ def process_donation_event(data):
...
+                    "server_name": settings.STATSD_PREFIX,
+                    "fail_reason": fail_reason,
+                },
+            )
+            send_mail(
+                email_title,
+                body,
+                "noreply@mozilla.com",
+                [settings.DONATE_NOTIFY_EMAIL],
+            )
```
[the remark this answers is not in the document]
stevejalim: Is this triggered maybe via an API call from https://donate.mozilla.org/en-US/, which is run by MoFo?

basket/news/backends/ctms.py
```diff
@@ -873,4 +873,4 @@ def ctms_interface():
         return None
 
 
-ctms = CTMS(ctms_interface(), not settings.SFDC_ENABLED)
+ctms = CTMS(ctms_interface(), True)
```
stevejalim, line 876: Nit: Just to save the grandkids having to see what `True` refers to

```suggestion
ctms = CTMS(ctms_interface(), is_primary=True)
```

basket/news/tasks.py
```diff
@@ -328,19 +322,6 @@ def fxa_direct_update_contact(fxa_id, data):
...
-            # couldn't find the fxa_id and tried to create a record but doesn't
-            # have the required data to do so. We can drop this one.
-            return
-        else:
-            # otherwise it's something else and we should potentially retry
-            raise
-    except SFDCDisabled:
-        pass
-
     basket_data = from_sfdc(data)
```
stevejalim, line 325: Can we lose `from_sfdc` now, too, or is that going in a later pass?

basket/news/tasks.py
```diff
@@ -1314,122 +1132,12 @@ def amo_compress_categories(categories):
...
-                    "AMO_Contact_ID__c": user["id"],
-                },
-            )
-        except sfapi.SalesforceMalformedRequest as e:
-            if e.content[0]["errorCode"] == "DUPLICATE_VALUE":
-                # dupe error, so we don't need to do this again
-                pass
-            else:
-                raise e
+    pass
```
stevejalim, line 1140: I agree with the `pass` here, but also wonderning if a deprecation warning might be handy, to remind us that the AMO-related behaviour is a no-op?

basket/settings.py
```diff
@@ -222,23 +218,6 @@ def path(*args):
     "watchman.checks.databases",
 )
 
-# Salesforce.com
```
stevejalim, line 225: There are Salesforce-related secrets in referenced in https://github.com/mozmeao/basket-config/blob/master/mozmeao-fr/basket-prod/web-deploy.yaml, so they prob should be removed too - and once that's done and released, we can cull the actual secrets from https://github.com/mozmeao/basket-private-config

stevejalim (approved):
Looks sensible to me, and always great to see more code being removed than added 😄 

Threw in some thoughts/ideas, but nothing blocking, I think. r+

basket/news/tasks.py
robhudson: I'm going to work on that next. Basically bypass the `from_sfdc` by creating the `data` in the right format to begin with. Then we can remove it.

basket/settings.py
robhudson: Good plan, thanks for the pointer.

basket/news/tasks.py
robhudson: Sure, what would that look like exactly? I could raise an exception? I'll scan other parts of the code and maybe there is already a pattern for things like this.

basket/news/tasks.py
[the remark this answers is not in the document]
robhudson: @pmac Do you have some historical perspective here? Would this be something we need more code around or can safely remove?

basket/news/tasks.py
stevejalim: I think raising a `DeprecationWarning` might be the way forward, if I'm right that it won't actually blow up things, just shout loudly https://docs.python.org/3.9/library/exceptions.html#DeprecationWarning

```from warnings import warn

warn('This is a deprecated thing that needs a better message than this example', DeprecationWarning, stacklevel=2)
```

basket/news/tasks.py
robhudson: This is now removed.

basket/news/tasks.py
```diff
@@ -329,34 +317,20 @@ def fxa_direct_update_contact(fxa_id, data):
...
+        ctms.update_by_alt_id("fxa_id", fxa_id, data)
     except CTMSNotFoundByAltIDError:
         # No associated record found, skip this update.
         pass
 
 
 @et_task
 def fxa_delete(data):
-    fxa_direct_update_contact(data["uid"], {"FxA_Account_Deleted__c": True})
+    fxa_direct_update_contact(data["uid"], {"fxa_deleted": True, "newsletters": []})
```
pmac, line 328: This doesn't unsubscribe people from newsletters right?

basket/news/utils.py
```diff
@@ -366,35 +360,9 @@ def get_user_data(
...
-            "Email service provider auth failure",
-            error_code=errors.BASKET_EMAIL_PROVIDER_AUTH_FAILURE,
-            status_code=500,
-        )
-    except SFDCDisabled:
-        sfdc_enabled = False
-        user = {}
+    user = {}
 
     if not user.get("email_id"):
```
pmac, line 365: We no longer need this `if` statement, but there's enough churn as it is, so IMO it's fine to leave it. Might confuse someone in future though.

basket/settings.py
pmac: We also need to be sure that we're using the proper sandbox variable, we may well be relying on `EXACTTARGET_USE_SANDBOX` value to set `USE_SANDBOX_BACKEND` for deployments.

pmac (approved):
r+wc. Just want to make sure that the call in the FxA deleted task doesn't unsubscribe users from everything, as that is not necessarily what they intended by deleting FxA. Though we do need to unsub them from the Accounts Tips newsletter in that case, which is #899

basket/news/tasks.py
robhudson: Definitely a good call out here. The changes here are attempting to reproduce what the `from_sfdc` code did with the `data` passed in. The tests don't cover any newsletter mapping via this call, but I will dig deeper here.

basket/news/utils.py
robhudson: I'm of the mind that if there's lots of churn, might as well get all the churn in at once so it's one commit of churn.

basket/news/tasks.py
robhudson: To remove the subscription to all newsletters requires a special string, sending the empty list is basically a no-op. See: https://github.com/mozilla-it/ctms-api/blob/main/tests/unit/test_api_patch.py#L440

basket/news/tasks.py
robhudson: I removed the AMO sync code

basket/news/utils.py
```diff
@@ -362,47 +362,45 @@ def get_user_data(
...
-                    error_code=errors.BASKET_NETWORK_FAILURE,
-                    status_code=400,
-                )
-        except CTMSNotConfigured:
+    ctms_user = None
+    try:
+        ctms_user = ctms.get(
+            token=token,
+            email=email,
+            amo_id=amo_id,
```
pmac, line 370: don't need `amo_id` any more either right?

robhudson: Probably, I'll check out what's calling this and clean that up as well. If no AMO stuff is happening it would be nice to simplify. 👍

PR #934, opened 2022-11-29
Fix #918: Only validate API key on SUBSCRIBE

### LINKED ISSUE #918: Remove API key requirement when unsubscribing from newsletters
As part of https://github.com/mozilla/bedrock/issues/11149 we'd like to move as much of the newsletter preference center functionality to the front-end. In order to facilitate, we will need to remove the API key requirement when unsubscribing from newsletters (e.g. in the instance a newsletter is marked as private). 

After some initial discussion, our current thinking is that removing the API key requirement here should be OK, since the worst someone could do is unsubscribe an email address from newsletters they have previously signed up to. In order to do that, they would also need a user's token, which would already give them access to do so anyways via the web site.

robhudson:


basket/news/views.py
```diff
@@ -891,7 +891,7 @@ def update_user_task(request, api_call_type, data=None, optin=False, sync=False)
                     400,
                 )
 
-            if api_call_type != UNSUBSCRIBE and nl in private_newsletters:
+            if api_call_type == SUBSCRIBE and nl in private_newsletters:
```
pmac, line 894: well... I don't think this is everything. We also need to ensure that a private newsletter is dropped from the list when the type is `SET` and they're not already subscribed to it.

pmac (changes requested):

basket/news/views.py
robhudson: Okay. Maybe I misunderstood.

There's 2 things here:
1. We don't know the current set of newsletters the user is subscribed to in the view. So if we need that, we'll need to make a call out to CTMS to get that info, but it was my understanding we'd like to avoid that.
2. We mentioned in our chat that this shouldn't be possible since the form won't show private newsletters to users, but since this is an API I can understand wanting to protect against that, but if so, perhaps we can handle that in the task and drop any private newsletters there.

pmac: correct. we discussed doing #2, just dropping the private newsletter from the list if the user is not already subscribed, and doing so in the task where we already have the user data.

pmac: and yes, while the form should not allow this, the API is open and so we should never rely on the form for validation.

robhudson: Updated to include the above.

pmac (approved):
🥇

PR #936, opened 2022-12-01
Python dependency bumps

robhudson:


stevejalim (approved):
r+ wc

robhudson:
I was intending for a general update to most deps, but the dev ones were the only pinned ones that needed updating.

PR #965, opened 2023-02-01
Fix #930: Mask email in user data requests

### LINKED ISSUE #930: Mask Email Addresses in User Data Requests
Unless a request comes in with a valid API key we should mask the email address in the returned user data. This would prevent anyone learning a users full email if they learned their token or email pref center URL. Since one can't update their email from the pref center displaying the full email isn't needed and it would improve privacy if it wasn't displayed unmasked.

robhudson:
Mask emails for user data requests if no API key provided.

### DIFF
--- a/basket/news/views.py
+++ b/basket/news/views.py
@@ -657,7 +657,8 @@ def user(request, token):
         return update_user_task(request, SET, data)
 
     get_fxa = "fxa" in request.GET
-    return get_user(token, get_fxa=get_fxa)
+    masked = not has_valid_api_key(request)
+    return get_user(token, get_fxa=get_fxa, masked=masked)
 
 
 @require_POST
@@ -815,7 +816,8 @@ def lookup_user(request):
             400,
         )
 
-    if email and not is_authorized(request, email):
+    authorized = is_authorized(request, email)
+    if email and not authorized:
         return HttpResponseJSON(
             {
                 "status": "error",
@@ -833,7 +835,7 @@ def lookup_user(request):
 
     try:
         user_data = get_user_data(
-            token=token, email=email, get_fxa=get_fxa, masked=not email
+            token=token, email=email, get_fxa=get_fxa, masked=not authorized
         )
     except NewsletterException as e:
         return newsletter_exception_response(e)
--- a/basket/news/views.py
+++ b/basket/news/views.py
@@ -754,9 +754,9 @@ def newsletters(request):
 
 @never_cache
 def lookup_user(request):
-    """Lookup a user in Exact Target given email or token (not both).
+    """Lookup a user in CTMS given email or token (not both).
 
-    To look up by email, a valid API key are required.
+    To look up by email, a valid API key is required.
 
     If email and token are both provided, an error is returned rather
     than trying to define all the possible behaviors.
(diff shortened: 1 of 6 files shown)


stevejalim (approved):
LGTM!

Nit, not blocking: Worth updating the `get_user_data` docstring to mention the masking, too?

robhudson:
@alexgibson Just wanted to bring this to your attention in case it affects the newsletter front-end.

basket/news/views.py
```diff
@@ -833,7 +833,9 @@ def lookup_user(request):
             return invalid_email_response()
 
     try:
-        user_data = get_user_data(token=token, email=email, get_fxa=get_fxa)
+        user_data = get_user_data(
+            token=token, email=email, get_fxa=get_fxa, masked=not email
```
pmac, line 837: I think we should enable returning the unmasked email when they use an API key regardless of whether they've requested via email or token. 

We also need to do this to the `user()` view since that is the one that bedrock uses to fetch user data.

alexgibson:
It’s possible I’m misunderstanding the scope of this change, but wouldn’t this break unsubscribing, where we need to pass the user’s email? https://basket.readthedocs.io/newsletter_api.html#news-unsubscribe

pmac:
I thought unsubscribe was all handled through the email prefs center which POSTs to `/news/user/<token>/`?

pmac:
If we do use `/news/unsubscribe/` then you're right that we'd need to make sure it works with token as well.

pmac:
actually, looking at the basket code, I think the docs are just wrong, the request should be going to `/news/unsubscribe/<token>/` and it doesn't require email, only `newsletters` or `optout`.

alexgibson:
@pmac I took a look and we do post to `/news/unsubscribe/<token>/`, however we do also pass `email` along with `optout`. If this isn't actually needed, then we'll need to remove that from the management page. I'll file an issue to do so.

basket/news/views.py
robhudson: Reading the `lookup_user` code, an API key is only validated when requested with the `email` param. Also, the `user` view doesn't support user lookup by email, just token, so it makes sense to always return masked emails.

Unless you're saying you want a change in the API to allow user lookups with token + API key?

pmac: Maybe I am. I'm thinking that this will add to the current situation. Currently to use `lookup-user` with email requires an API key because that's privileged info. And now I think we should add to that such that if you send a valid API key along with a request to either of these endpoints you get the full email address back, but otherwise you only get a redacted one. Does that make sense?

robhudson: Yeah, makes sense. So API key plus either token or email replies with unmasked email. I can start working towards that in this PR.

robhudson:
Updated to add the ability to get unmasked emails with token lookups + API key. I'll do a pass on the docs to see if there are any things needing updated to match these changes. Otherwise this is ready for re-review.

alexgibson:
Bedrock PR to remove use of email from management page POST requests: https://github.com/mozilla/bedrock/pull/12703

pmac (approved):
💯

PR #981, opened 2023-02-22
Flip the default of email masking

robhudson:
When deployed, this caused an issue since `get_user_data` is used internally and sends the output to acoustic, which was then masked. This update default to masking off and only masks in the views fetching user data.

### DIFF
--- a/basket/news/tasks.py
+++ b/basket/news/tasks.py
@@ -612,7 +612,7 @@ def upsert_contact(api_call_type, data, user_data):
 
     if send_confirm and settings.SEND_CONFIRM_MESSAGES:
         send_confirm_message.delay(
-            user_data["email"],
+            data["email"],
             token,
             update_data.get("lang", user_data.get("lang", "en-US")),
             send_confirm,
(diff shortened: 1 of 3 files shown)


pmac (changes requested):
Is it possible to add anything extra to tests or some extra tests to ensure that masked email doesn't make it to the acoustic backend again?

robhudson:
> Is it possible to add anything extra to tests or some extra tests to ensure that masked email doesn't make it to the acoustic backend again?

I looked into that but the tests I checked the data from `get_user_data` was mocked and unmasked. So the data being tested is valid since `get_user_data` won't mask by default anymore.

I could add a test that will fail if `get_user_data` flips the default. Maybe a future programmer will want to do that and the test can indicate why we coded it that way.

pmac:
Right. Makes sense. It's probably fine then. We very rarely mess with this stuff. I think maybe a good note in the doc string about the False default would be nice though.

robhudson:
I was already writing the tests. I just put in a couple to test the default value of the kwarg but this method could use more tests in the future.

pmac (approved):
🍔

PR #1007, opened 2023-04-12
Add isort

robhudson:


.pre-commit-config.yaml
```diff
@@ -4,12 +4,12 @@ repos:
       hooks:
           - id: black
             language_version: python3.9
-    - repo: https://github.com/asottile/add-trailing-comma
```
pmac, line 7: Is this functionality part of isort or something?

pmac (approved):
🌴

.pre-commit-config.yaml
robhudson: No, but I was trying to be more consistent with bedrock which doesn't have it. If it is preferred I can keep it in there and consider adding it to bedrock.

pmac: I think it used to be in bedrock. Maybe black does it now? I'm not sure.

PR #1008, opened 2023-04-12
Run unit tests using GitHub Actions

robhudson:


.circleci/config.yml
```diff
@@ -1,20 +0,0 @@
...
-      - image: cimg/python:3.9
-    steps:
-      - checkout
-      - setup_remote_docker:
-          docker_layer_caching: false
-          version: 20.10.11
-      - run:
-          name: Setup
-          command: |
-            echo "export GIT_COMMIT=${CIRCLE_SHA1}" >> $BASH_ENV
```
stevejalim, line 14: I think you might need to do something like this for the GHA so that the build has it

eg add a new step with

`echo "GIT_COMMIT=${CIRCLE_SHA1}" >> $GITHUB_OUTPUT`

.github/workflows/pull_request_tests.yml
```diff
@@ -0,0 +1,18 @@
...
+  test-python:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.9"  # matches current Python in production
+      - name: "Run Python tests (on Docker)"
+        run: |
+            make clean test-image
```
stevejalim, line 18: Sorry to ask: should this be `make build-ci` and `make test-ci` ?

stevejalim (commented):
Coupla questions

.github/workflows/pull_request_tests.yml
robhudson: `make test-ci` triggers the `make build-ci`

.circleci/config.yml
robhudson: Yeah this may be needed to see the commit hash. I'll check it out. Thanks for the callout.

stevejalim (approved):
Ignore my previous review - looks like it works a treat! R+

.circleci/config.yml
robhudson: Looks like this is working without this being explicitly set. See [set_git_env_vars.sh](https://github.com/mozmeao/basket/blob/master/docker/bin/set_git_env_vars.sh#L5)

PR #1014, opened 2023-04-14
Fix `KeyError` on vpn waitlist platform, which is no longer required

robhudson:


basket/news/backends/ctms.py
```diff
@@ -119,7 +119,8 @@ def from_vendor(contact):
                 if wl_name == "vpn":
                     wl_name = VPN_NEWSLETTER_SLUG
                     data["fpn_country"] = waitlist["fields"]["geo"]
-                    data["fpn_platform"] = waitlist["fields"]["platform"]
+                    if platform := waitlist["fields"].get("platform"):
```
pmac, line 122: image

pmac (approved):
🦭

Issue #1023, opened 2023-04-21
"Manage your Email Preferences" page is throwing an error


LuminitaB:
**FxA version:**
1.256.3

**Prerequisites:**
Have a Fxa account;

**Steps to reproduce:**

1. Navigate to https://accounts.stage.mozaws.net/ and login using an existing account;
2. Click on the "Email communications" option displayed in the navigation menu;
3. Click on "Sign in" button;

**Expected result:**
The user is taken to the Manage your Email Preferences page and the options are successfully displayed;

**Actual result:**
The user is taken to the Manage your Email Preferences page but an error is displayed;

Notes:
The issue reproduces intermittently and/or for specific accounts. I was able to constantly reproduce with: lburuian+fxaprod@mozilla.com | Test1234!

**Another scenario:**
1. For an account that works, make some selections for email preferences and save
2. Go back to preferences and select to remove from all subscriptions and save
3. Go back to check that all options have been deselected
I get the same error!

![image](https://user-images.githubusercontent.com/52028275/233591005-3455598b-6b05-427e-9b7a-298f4241a0f1.png)

alexgibson:
Possible related to https://github.com/mozmeao/basket/issues/928

pmac:
In my testing with the `another scenario` method I don't see what you're seeing Alex. I see all 200 responses from basket, but I do see that the user data does not come back with a `newsletters` key. Perhaps this is a change in basket.

pmac:
We just landed a fix that has fixed the issue Alex and I were able to reproduce on staging. I'll get it to prod shortly.

LuminitaB:
Verified as fixed on Train 1.257.1
We did not encounter the error anymore.
Thank you.

PR #962, opened 2023-01-25
Ref #952: Waitlists as relationships

leplatrem:
Ref #952

In this pull-request, we propose an implementation that leverages the new waitlists format in CTMS (see https://github.com/mozilla-it/ctms-api/pull/492)

The main idea is to be able to onboard waitlists with any arbitrary field, without having to redeploy Basket. 
The waitlists fields validation is fully managed by CTMS.

Basically, with this PR, when the following data is posted:
```
newsletters=vpn-bundle-waitlist
vpn_bundle_currency=eur
vpn_bundle_pack=gold
source_url=http://mozilla.org
```
The following payload is sent to CTMS:
```
"waitlists": [
  {
     "name": "vpn-bundle",
     "subscribed": True,
     "source": "http://mozilla.org",
     "fields": {"currency": "eur", "pack": "gold"},
  },
]
```

Onboarding a new waitlist would consist in:
1. Add new newsletter `abc-xyz-waitlist` with the `is_waitlist` flag set to True (name suffix is only by convention)
2. Create a form that sends `abc-xyz-waitlist` in the `newsletter` field, and optionally sends arbitrary data prefixed with `abc_xyz_`

### Documentation

* https://github.com/mozilla-it/ctms-api/blob/main/docs/adrs/waitlists_relationships.md
* https://mozilla-hub.atlassian.net/browse/CTMS-136
* old proposition about Basket inbox endpoint https://github.com/mozilla-it/ctms-api/pull/371

### Next steps

- [x] global approach is validated (see Rob comment)
- [x] unit tests pass
- [x] verify that vpn waitlist newsletter is called ~~`"fpn"`~~ `"guardian-vpn-waitlist"` in Basket database
- [x] verify that `newsletters` field is always sent in existing waitlist subscription forms (✅ relay waitlists, ✅ guardian vpn)
- [x] behaviour is validated using integration tests https://github.com/mozilla-it/ctms-api/pull/570
- [ ] this pull request is approved (review by commit should be easier)
- [x] CTMS new waitlist API is deployed and validated in production (it's retro compatible)
- [ ] this pull request is merged and deployed

PR #1034, opened 2023-04-24
Ensure basket always returns a list of newsletters with user data

pmac:
Even if said list is empty


pmac:
My theory on #1023 is that the page is not expecting the response from basket to have no "newsletters" key in the data. This change ensures that it will return an empty list if there are no subscriptions.

robhudson (approved):
Great sleuthing. 🍏

pmac:
Looks like the change here is the culprit https://github.com/mozmeao/basket/pull/962/files#diff-d6aadfae5982cb77055ddff406cd982f1dce929c139d9dd562ca488f0b9ad568L106 that is subtle, but a side effect is that it won't necessarily always return a `newsletters` key in the data even if CTMS sends an empty one to us.

PR #1036, opened 2023-04-27
Fix #1035: Add content-type header for Acoustic Transact requests

### LINKED ISSUE #1035: Add content-type header for Acoustic Transact requests
Transact customers using Oauth for submissions should add header: 

`Content-Type : text/plain` or 
`Content-Type : text/xml`

robhudson:


pmac (approved):
🍎

PR #1048, opened 2023-05-10
Dependency bumps

robhudson:


basket/news/migrations/0027_alter_newsletter_requires_double_optin.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 3.2.19 on 2023-05-10 17:06
```
pmac, line 1: Why is this necessary?

pmac (approved):
Looks good. Not sure why there is a migration, but assuming it's needed this is good to go ✅

basket/news/migrations/0027_alter_newsletter_requires_double_optin.py
robhudson: String change in `help_text` to add a space after the period. Kinda lame but 🤷‍♂️ .

stevejalim: Agreed @robhudson  - I'm not a fan Django needing a new migration when tweaks to, essentially, Python strings have to happen. In the past, when am sure it's a no-op change in terms of real DB schema, I've dug out the most recent migration to touch the field and edited that old migration to match, so that Django no longer sees the change. It's a little risky because it might mess with a git-bisect if someone needs to dig into history, but it can work if the planets are appropriately aligned

PR #1056, opened 2023-05-16
Migrate to using Django's built-in `JSONField`.

robhudson:


basket/news/migrations/0018_commonvoiceupdate.py
```diff
@@ -31,7 +29,7 @@ class Migration(migrations.Migration):
                         editable=False,
                     ),
                 ),
-                ("data", jsonfield.fields.JSONField(default=dict)),
+                ("data", models.JSONField(default=dict)),
```
robhudson, line 32: Had to edit these old ones due to import errors b/c the package was removed. But it should only come into play for new installs or tests.

basket/news/migrations/0028_django_jsonfield.py
```diff
@@ -0,0 +1,37 @@
+# Generated by Django 3.2.19 on 2023-05-16 22:50
```
pmac, line 1: With the editing of the earlier migrations is this one still necessary?

pmac (approved):
r+wc

basket/news/migrations/0028_django_jsonfield.py
pmac: squash++

PR #1069, opened 2023-06-01
Transition basket tasks to use rq

robhudson:
Still some work to be done here...

- [x] Fully test tasks locally
- [x] Update documentation (if needed)
- [x] Add snitch management command
- [x] Add snitch k8s `[CronJob](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/)`
- [x] Make derived settings from `REDIS_URL` also Django settings, rather than env variables.


basket/base/tests/test_rq_utils.py
```diff
@@ -0,0 +1,239 @@
...
+        args = ["arg1"]
+        kwargs = {"arg2": "foo"}
+        job = failing_job.delay(*args, **kwargs)
+
+        worker = get_worker()
+        worker.work(burst=True)  # Burst = worker will quit after all jobs consumed.
+
+        assert job.is_failed
+
+        # TODO: Determine why the `store_task_exception_handler` is not called.
```
stevejalim, line 97: Is this a blocker or OK as a known-unknown?

stevejalim (approved):
Looking good to me. Some questions but curiosity rather than blockers

basket/base/management/commands/rqworker.py
```diff
@@ -0,0 +1,37 @@
+import sys
+
+from django.core.management.base import BaseCommand
+
+from basket.base.rq import get_worker
+
+
```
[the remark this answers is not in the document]
robhudson: Yes, I chose to simplify and not use `django_rq` since it adds a lot of indirection I wasn't too crazy about. But looking at rq's [`rqworker`](https://github.com/rq/rq/blob/master/rq/cli/cli.py#L198) CLI makes me curious if I need a bit more in here.

basket/base/rq.py
```diff
@@ -0,0 +1,230 @@
...
+            # Force retries to abort.
+            # Since there's no way to abort retries at the moment in RQ, we can set the job `retries_left` to zero.
+            # This will retry this time but no further retries will be performed.
+            job.retries_left = 0
+
+        return
+
+    if job.is_failed:
+        statsd.incr(f"{task_name}.retry_max")
+        statsd.incr("news.tasks.retry_max_total")
```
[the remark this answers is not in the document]
robhudson: Yeah, we'd get a count of how many times a specific task hits the max retries, as well as how many times all jobs hit the max retries.

basket/base/tests/test_rq_utils.py
robhudson: I added more tests so there's 100% coverage, it's just broken into separate tests. But I'd like to solve this riddle at some point. So I don't consider it a blocker, just a hard thing to test.

basket/base/rq.py
```diff
@@ -0,0 +1,230 @@
...
+# don't propagate and don't retry if these are the error messages
+IGNORE_ERROR_MSGS = [
+    "INVALID_EMAIL_ADDRESS",
+    "InvalidEmailAddress",
+    "An invalid phone number was provided",
+    "No valid subscribers were provided",
+    "There are no valid subscribers",
+    "email address is suppressed",
+    "invalid email address",
+]
```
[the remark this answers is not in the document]
pmac: Some of these are probably legacy too. Because we own the CTMS client fully we could probably create and throw specific exceptions and catch and bubble the ones we want that way. Though that would be harder for Acoustic.

basket/base/rq.py
```diff
@@ -0,0 +1,230 @@
...
+    """
+    Get a Redis connection.
+
+    Expects a URL including the db, or defaults to `os.environ["RQ_URL"]`.
+
+    Call example:
+        get_redis_connection("redis://localhost:6379/0")
+
+    """
+    global _REDIS_CONN
```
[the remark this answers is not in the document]
pmac: it's the kind of thing I've always done too. Seems good to me.

basket/base/rq.py
```diff
@@ -0,0 +1,231 @@
...
+
+    if url is None:
+        url = os.environ.get("RQ_URL")
```
pmac, line 60: any reason you're not using Django settings here?

basket/base/rq.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+def log_timing(job):
+    if start_time := job.meta.get("start_time"):
+        total_time = int(time() - start_time) * 1000
+        statsd.timing(f"{job.meta['task_name']}.duration", total_time)
+
+
+def rq_on_success(job, connection, result, *args, **kwargs):
+    # Don't fire statsd metrics in maintenance mode.
```
pmac, line 142: why not?

basket/base/tests/test_rq_utils.py
```diff
@@ -0,0 +1,239 @@
...
+class TestRQUtils(TestCase):
+    @override_settings(RQ_MAX_RETRIES=10)
+    def test_rq_exponential_backoff(self):
+        """
+        Test that the exponential backoff function returns the correct number
+        of retries, with a minimum of 60 seconds between retries.
+        """
+        with patch("basket.base.rq.random") as mock_random:
+            mock_random.randrange.side_effect = [2 * 2**n for n in range(settings.RQ_MAX_RETRIES)]
+            self.assertEqual(rq_exponential_backoff(), [60, 60, 60, 60, 60, 64, 128, 256, 512, 1024])
```
pmac, line 33: That's only a total of around 40 minutes. Our current backoff can go on for days. Is this right, or can we extend it further?

basket/settings.py
```diff
@@ -29,6 +27,7 @@ def path(*args):
 
 
 DEBUG = config("DEBUG", default=False, cast=bool)
+TESTING = config("TESTING", default=False, cast=bool)
```
pmac, line 30: What is this setting for? Only when unittests are being run, or something else?

bin/run-worker.sh
```diff
@@ -1,7 +1,3 @@
 #!/bin/bash -ex
 
-exec newrelic-admin run-program celery -A basket.news worker \
-                                       -P "${CELERY_POOL:-prefork}" \
-                                       -l "${CELERY_LOG_LEVEL:-warning}" \
-                                       -c "${CELERY_NUM_WORKERS:-4}" \
-                                       -Q celery,snitch
+exec python manage.py rqworker --with-scheduler
```
pmac, line 3: we dropping new relic?

pmac (commented):
This is looking fantastic so far! Can't wait to get it landed. I think it's nearly there.

basket/base/management/commands/rqworker.py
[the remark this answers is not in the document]
robhudson: I did review the `cli/rqworker` and I think we're good. We're picking up a lot of the same defaults just by using our `get_worker` and default kwargs stuff.

basket/base/rq.py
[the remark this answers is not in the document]
robhudson: That seems worthwhile. I could create an issue for it that we can add post-merge.

basket/base/rq.py
robhudson: Not really, other than that's the pattern that already existed in the settings file. We're creating the `RQ_URL` env variable from the `REDIS_URL`. But we could also make it a django setting rather than an env variable. I'll add it to the task in the PR to clean this up.

basket/base/tests/test_rq_utils.py
robhudson: Ooh, good catch. In converting this I got mixed up b/c the celery code is working against minutes whereas I'm working against seconds. So I need to increase the factor here from `2 * 2**n` to `120 * 2**n` to be more inline with what we had with celery.

With that change this would be:
`[120, 240, 480, 960, 1920, 3840, 7680, 15360, 30720, 61440]`
without the jitter which looks better.

Thanks for checking my math.

basket/settings.py
robhudson: Yes just for unit tests

bin/run-worker.sh
robhudson: Nope, missed that. My eyes were searching for `celery`. I'll make a note to add that back in.

basket/base/rq.py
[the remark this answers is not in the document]
robhudson: https://github.com/mozmeao/basket/issues/1081

basket/base/rq.py
robhudson: In maintenance mode we're going to queue the task in the db rather than do the task, so we'd get timings for doing the db insert rather than timings for the task, which would confuse the timings. And a "success" for the db insert isn't the same as a "success" for the task.

basket/settings.py
pmac: Any reason to have it be a setting like this rather than be detected like we did at the bottom of the file? Or is there some mechanism that will set a `TESTING` env var when we're running tests?

robhudson: I added it to `docker/envfiles/test.env` and at one point I thought I added it to the pytest env settings for running tests outside of Docker, but I don't see that at the moment. I found the settings changes at the bottom of the file when run with certain commands to be less explicit and "hidden" -- I hadn't actually ever noticed that before, as compared to keeping these settings near each other with the `TESTING` conditional. If you prefer it like before I can change it back.

robhudson: I remember adding `pytest-env` and getting that to work, but then I removed it. The reason why is I have never actually run basket outside of Docker due to the `mysql-client` requirement and not wanting to install mysql on my host machine when we have a fully capable docker environment. And since this PR is adding a Redis requirement also, I don't see any reason to support running outside of a Docker environment.

pmac: True. If it works it's okay with me. I was just thinking something more like:

```python
TESTING = False
if sys.argv[0].endswith("py.test") or sys.argv[0].endswith("pytest") or (len(sys.argv) > 1 and sys.argv[1] == "test"):
     TESTING = True
```

Just to avoid having tests fail if we forget to set an env var during tests. While running the system does require everything, just running unit tests shouldn't. I tend to run them locally.

basket/settings.py
```diff
@@ -70,6 +61,15 @@ def path(*args):
     }
 DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
 
+# CACHE_URL and RQ_URL are derived from REDIS_URL.
+REDIS_URL = config("REDIS_URL", None)
+if REDIS_URL:
+    REDIS_URL = REDIS_URL.rstrip("/0")
+    HIREDIS_URL = REDIS_URL.replace("redis://", "hiredis://")
+    # use redis for cache and rq.
+    os.environ["CACHE_URL"] = HIREDIS_URL + "/" + config("REDIS_CACHE_DB", "1")
```
pmac, line 70: I'd have to look, but I think the reason we set these in the env rather than just keeping them settings is that celery would auto setup some things from the env in ways that wasn't easy via django settings. I don't think we have that same situation here do we? I could also be misremembering.

basket/settings.py
robhudson: I added this in and changed the setting name to `UNITTEST`.

PR #1116, opened 2023-07-19
Clean up unused settings and middleware

robhudson:


basket/news/middleware.py
```diff
@@ -51,39 +34,3 @@ def __call__(self, request):
         response = self.get_response(request)
         response["X-Backend-Server"] = self.backend_server
         return response
-
-
-def is_ip_address(hostname):
-    return bool(IP_RE.match(hostname))
-
-
-class EnforceHostnameMiddleware(object):
```
pmac, line 60: Is this no longer in use?

basket/news/middleware.py
```diff
@@ -1,24 +1,7 @@
...
-from django.http.request import split_domain_port
 
 from django_statsd.clients import statsd
 from django_statsd.middleware import GraphiteRequestTimingMiddleware
-from mozilla_django_oidc.middleware import SessionRefresh
-
-IP_RE = re.compile(r"^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$")
-
-
-class OIDCSessionRefreshMiddleware(SessionRefresh):
```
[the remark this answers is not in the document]
robhudson: I flipped it to remove the custom middleware. Instead of saying "OIDC only the admin", I'm saying "exclude everything but the admin". It's effectively the same but just means we need to exclude new URLs we add. I added a note in the `urls.py` file for that.

basket/news/middleware.py
robhudson: I chatted with Brett and it appears to not be used.

pmac (approved):
💯

PR #1139, opened 2023-08-14
Switch to the datadog backend

robhudson:
This reverts commit 33ee26c264915c76e1613ed254e8aac2704fcdaa.


robhudson:
Merging to test on dev/stage. Assuming data flows through the same, I'll have further commits to optimize the use of the metrics calls.

PR #1142, opened 2023-08-16
Optimize metrics calls via tagging

robhudson:
Using tags to add data and letting the metric name be less specific. With this approach we can select the metric and group by the tags to show charts in Grafana.

A nice example from socorro:
https://earthangel-b40313e5.influxcloud.net/d/LysVjx8Zk/socorro-prod-megamaster-remix?orgId=1&refresh=1m&from=now-6h&to=now&viewPanel=100

### DIFF
--- a/basket/base/middleware.py
+++ b/basket/base/middleware.py
@@ -16,19 +16,7 @@ class HostnameMiddleware(MiddlewareMixin):
         return response
 
 
-class MetricsStatusMiddleware(MiddlewareMixin):
-    """Send status code counts to statsd"""
-
-    def process_response(self, request, response):
-        metrics.incr(f"response.{response.status_code}")
-        return response
-
-    def process_exception(self, request, exception):
-        if not isinstance(exception, Http404):
-            metrics.incr("response.500")
-
-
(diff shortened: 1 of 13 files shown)


pmac:
This all looks good. A question came up recently about whether a change had affected the rate at which people subscribed to a particular newsletter. Is it possible with tagging to add the newsletter ID to subscription stats?

robhudson:
> This all looks good. A question came up recently about whether a change had affected the rate at which people subscribed to a particular newsletter. Is it possible with tagging to add the newsletter ID to subscription stats?

Yes! I like that. We can definitely have a successfully subscribed metric and have the newsletter ID as a tag so we can filter or group by them.

pmac:
Great! Wanna do that as a follow-up or add it to this PR?

pmac (approved):
👍

robhudson:
> Great! Wanna do that as a follow-up or add it to this PR?

Follow-up commit. I took a quick look. It's a bit more complex since there's multiple newsletters and they could be subscribing or updating their subscriptions or even unsubscribing, and it happens in a task. And we're just asking CTMS to update the records but we can still signal a metric to show we requested it.

PR #1162, opened 2023-09-05
Have Sentry ignore common but non-actionable errors

robhudson:


stevejalim (approved):

basket/settings.py
```diff
@@ -287,8 +287,18 @@ def get_default_gateway_linux():
...
+    BrokenPipeError,
+    ConnectionResetError,
+)
+
 
 def before_send(event, hint):
+    if hint and "exc_info" in hint:
+        exc_type, exc_value, tb = hint["exc_info"]
+        if isinstance(exc_value, SENTRY_IGNORE_ERRORS):
+            return None
```
pmac, line 300: I think you have to return `event` but we may need to check the docs.

robhudson: I borrowed this from bedrock, but returning null is the correct return value to ignore the event...

"It [`before_send`] receives the event object as a parameter, so you can use that to modify the event’s data or drop it completely (by returning null) based on custom logic and the data available on the event."

pmac: sorry, I failed to notice the bit below where it does return event after some other processing. Ignore me 🤦

PR #1198, opened 2023-10-27
Update fxa-error redirect

robhudson:


pmac (approved):
Looks good. We can push this out after bedrock is deployed.

PR #1207, opened 2023-10-31
Fix #1178: Re-send confirmation email if never confirmed

### LINKED ISSUE #1178: Ensure double-opt-in emails are sent if the user has not clicked to opt-in
We have reports that if a user has attempted to subscribe to a newsletter, but for whatever reason has not clicked the link in the opt-in email to opt-in, then attempts to subscribe again or to another newsletter that they won't receive another email with the opt-in link. I'm guessing this is for cases when a user attempts to subscribe to the same newsletter again since I believe basket will see that as a no-op since they're already subscribed technically. But if they don't have the opt-in then they are in limbo. So we should be checking that and sending the opt-in email to them again.

robhudson:


basket/news/tasks.py
```diff
@@ -381,6 +382,15 @@ def upsert_contact(api_call_type, data, user_data):
     else:
         ctms.update(user_data, update_data)
 
+    # In the rare case the user hasn't confirmed their email and is subscribing to the same newsletter, send the confirmation again.
+    if user_data and user_data.get("optin", False) is False and any(c not in update_data["newsletters"] for c in cur_newsletters):
```
pmac, line 386: I'm not following what `any(c not in update_data["newsletters"] for c in cur_newsletters)` is checking for. Shouldn't `parse_newsletters()` ensure that nothing in `cur_newsletters` is in `update_data["newsletters"]`? I think we're looking for the situation where `update_data["newsletters"]` is empty right?

robhudson: It's checking if there's any newsletters in `cur_newsletters` that aren't in the `update_data["newsletters"]` b/c `parse_newsletters()` removes them. But I'm not sure a simple empty check would work if they have a single currently subscribed newsletter and a new newsletter. In that specific case the newsletters data wouldn't be empty and we'd skip this block. But maybe that's okay?

pmac (approved):
💯 🎆

PR #1208, opened 2023-11-01
Initial set up for accepting petition signatures

robhudson:


basket/settings.py
```diff
@@ -144,10 +144,10 @@ def path(*args):
 MIDDLEWARE = (
     "allow_cidr.middleware.AllowCIDRMiddleware",
     "django.middleware.security.SecurityMiddleware",
+    "corsheaders.middleware.CorsMiddleware",
```
pmac, line 147: Why move this here? Doesn't that mean that all static media will get CORS headers? Is that the intent?

robhudson: I can revert it back since we won't use it for the petitions. Good catch. But yeah, the docs say to move it before common middleware so I was testing it out.

basket/petition/views.py
```diff
@@ -0,0 +1,32 @@
...
+# Doing this here to not affect the rest of basket.
+def _add_cors(response):
+    response["Access-Control-Allow-Origin"] = "https://open.mozilla.org"
+    response["Access-Control-Allow-Methods"] = "POST, OPTIONS"
+    return response
+
+
+def sign_petition(request):
+    if request.method == "POST":
+        form = PetitionForm(request.POST)
```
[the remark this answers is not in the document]
robhudson: @craigcook Any preference for form-urlencoded or sending JSON?

I was leaning towards XHR and JSON responses, but iiuc JS can submit POSTs either way and I thought form-urlencoded was kind of the default.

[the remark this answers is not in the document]
robhudson: Looks like we'll got with form urlencoded and return JSON.

basket/petition/views.py
```diff
@@ -0,0 +1,35 @@
...
+from django.conf import settings
+from django.http import JsonResponse
+from django.shortcuts import redirect
+
+from basket.petition.forms import PetitionForm
+
+
+# Doing this here to not affect the rest of basket.
+def _add_cors(response):
+    response["Access-Control-Allow-Origin"] = "https://open.mozilla.org"
```
stevejalim, line 12: Q: do we want this to be configurable? Probably not for now

stevejalim (approved):
r+ NICE

basket/petition/views.py
robhudson: Good catch. I can move it to settings. I already have one there so it makes sense for this too.

PR #1209, opened 2023-11-01
Add signatures.json

robhudson:


basket/petition/views.py
```diff
@@ -3,8 +3,10 @@
 from django.conf import settings
 from django.http import JsonResponse
 from django.shortcuts import redirect
+from django.views.decorators.cache import cache_page
```
stevejalim, line 6: I'm assuming that the basket prod setup has a real redis/memcached cache configured, yep?

By defualt, we're using locmem

https://github.com/mozmeao/basket/blob/6d8aa5582420906b6b37590056114a0c146894b0/basket/settings.py#L73-L84

stevejalim (approved):
🎸 r+wc

basket/petition/views.py
pmac: I believe we do. We can check with SRE. Either way this seems like a good thing.

pmac (approved):
🖊️

basket/petition/views.py
robhudson: `CACHE_URL` gets set if we have a `REDIS_URL`, so this should go into the Redis cache. Just above this block of code.

PR #1210, opened 2023-11-02
Add email confirmation flow

robhudson:


basket/petition/templates/petition/confirmation_email.txt
```diff
@@ -0,0 +1,6 @@
+Hi {{ name }}:
+
+Click this...
+
+Confirmation link:
+{{ confirm_link }}
```
robhudson, line 6: TODO: Update with the for realz text.

basket/petition/views.py
```diff
@@ -16,6 +20,24 @@ def _add_cors(response):
...
+
+    email_subject = "Please confirm your open.mozilla.org/letter signature"
+    email_body = render_to_string(
+        "petition/confirmation_email.txt",
+        {
+            "name": petition.name,
+            "confirm_link": confirm_link,
+        },
+    )
+    email_from = "no-reply@mozilla.org"
```
pmac, line 35: I think this is one of the approved from emails we can use. Did we confirm that? Last I used it SES required these be setup in their system.

pmac (approved):
Looks good to me. We can ship it once we get the final email messaging.

PR #1218, opened 2023-11-07
Move email confirmation sending to a task for retries

robhudson:
We are getting many SMTP authentication errors, even when the auth is correct, so the task will retry these.

### DIFF
--- a/basket/petition/views.py
+++ b/basket/petition/views.py
@@ -1,16 +1,15 @@
 import uuid
 
 from django.conf import settings
-from django.core.mail import send_mail
 from django.http import JsonResponse
 from django.shortcuts import redirect
-from django.template.loader import render_to_string
 from django.urls import reverse
 from django.utils.http import urlsafe_base64_decode, urlsafe_base64_encode
 from django.views.decorators.cache import cache_page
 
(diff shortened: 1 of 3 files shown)


basket/petition/views.py
```diff
@@ -1,16 +1,15 @@
...
 from django.http import JsonResponse
 from django.shortcuts import redirect
-from django.template.loader import render_to_string
 from django.urls import reverse
 from django.utils.http import urlsafe_base64_decode, urlsafe_base64_encode
 from django.views.decorators.cache import cache_page
 
 from basket.petition.forms import PetitionForm
 from basket.petition.models import Petition
+from basket.petition.tasks import send_email_confirmation
```
pmac, line 12: I don't see this tasks file. Maybe it wasn't added to the commit by mistake?

stevejalim (approved):
LGTM

basket/petition/views.py
```diff
@@ -59,7 +40,9 @@ def sign_petition(request):
         petition.save()
```
pmac, line 40: should we also catch a potential DB error here and specifically return an error, or just let it 500?

pmac (approved):
:shipit:

PR #1219, opened 2023-11-07
Add petition import management command

robhudson:


pmac (approved):
Looks good! Let's give it a shot!

basket/petition/models.py
```diff
@@ -18,7 +19,7 @@ class Petition(models.Model):
     referrer = models.TextField(blank=True)
     token = models.UUIDField()
     email_confirmed = models.BooleanField(default=False)
-    created = models.DateTimeField(auto_now_add=True)
+    created = models.DateTimeField(default=timezone.now)
```
stevejalim, line 22: Nice

stevejalim (approved):
r+

basket/petition/models.py
robhudson: I didn't realize the `auto_now_add` would completely ignore a datetime you pass in when creating a new record. So I changed it.

PR #1220, opened 2023-11-07
Various petition changes, grouped by commit

robhudson:


basket/petition/tests/test_views.py
```diff
@@ -54,7 +55,7 @@ def test_petition_post_success(client, mocker):
 
 @pytest.mark.django_db
 def test_petition_post_invalid(client, mocker):
-    mock_send_mail = mocker.patch("basket.petition.views.send_email_confirmation")
+    mock_send_mail = mocker.patch("basket.petition.models.send_email_confirmation")
```
stevejalim, line 58: I was expecting to see `"basket.petition.models.Petition.send_email_confirmation"` - is that patching def right?

basket/petition/tests/test_views.py
```diff
@@ -98,6 +98,27 @@ def test_petition_email_invalid(client, mocker):
...
+        "email": "thedude@example.com",
+        "title": "Dude",
+        "affiliation": "The Knudsens",
+    }
+    response = client.post(url, data=data)
+    assert response.status_code == 200
+    assert response.json() == {
+        "status": "error",
+        "errors": {
+            "__all__": "Database error",
```
stevejalim, line 116: Is this raised because of the emoji? Just wondering how it's triggered.

stevejalim (approved):
LGTM, with a couple of questions

r+wc

basket/petition/tests/test_views.py
robhudson: Ah, I'm patching the task which the model method calls.

basket/petition/tests/test_views.py
robhudson: Yeah an emoji will trigger it, or most any other database error, but the emoji is an easy test.

pmac (approved):
🥇

PR #1223, opened 2023-11-09
Add publish link to admin

robhudson:


pmac:
All looks good to me. This will only show on petitions admin page?

robhudson:
> All looks good to me. This will only show on petitions admin page?

Yep, it only shows up for that model, Petition.

PR #1258, opened 2023-12-30
Move settings to use django-configurations

robhudson:
I wanted to see what moving to django-configurations would look like. Overall I think it is an improvement. It's probably easier to view the settings file complete rather than the diff here.

Things I like:
- Simplifies overriding env vars in certain environments, e.g. `Testing`, or enabling sentry only in `Production` environments.
- Could easily allow for custom settings per deploy environment by adding `Dev`/`Stage`/`Prod` subclasses of `Production`. But, since we try to follow 12-factor, for now just using the env vars is all we needed here.
- Basket wasn't using the pattern of breaking settings across files and using `from base import *`, but django-configurations helps clean up that pattern using Python inheritance.

I'm interested in your thoughts with the consideration of using this across the MEAO org.

### DIFF
--- a/basket/base/tests/test__utils.py
+++ b/basket/base/tests/test__utils.py
@@ -4,11 +4,7 @@ import json
 from django.test.utils import override_settings
 
 from basket.base.utils import email_is_testing
-from basket.settings import (
-    SENSITIVE_FIELDS_TO_MASK_ENTIRELY,
-    SENSITIVE_FIELDS_TO_MASK_PARTIALLY,
-    before_send,
-)
+from basket.settings import SentryConfigurationMixin
 
 
 @override_settings(TESTING_EMAIL_DOMAINS=["restmail.net"], USE_SANDBOX_BACKEND=False)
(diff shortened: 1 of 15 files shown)


basket/base/tests/test__utils.py
```diff
@@ -4,11 +4,7 @@
 from django.test.utils import override_settings
 
 from basket.base.utils import email_is_testing
-from basket.settings import (
-    SENSITIVE_FIELDS_TO_MASK_ENTIRELY,
-    SENSITIVE_FIELDS_TO_MASK_PARTIALLY,
-    before_send,
-)
+from basket.settings import SentryConfigurationMixin
```
pmac, line 7: Should app (or test) code be using a mixin directly?

robhudson: Do you see an issue with it? It's only named "Mixin" to hint that it can be added as a base class for the other classes. The tests were importing from settings before for testing the sentry config, and it's doing the same after this changeset.

pmac: Only issue I see is that it's not testing what the code will be using (which I assume remains `settings`) and just looks in what's in a class, which could or could not be included in the actual settings. Would it be better to include this mixin with the testing config? Honestly, looking at these tests they seem mostly superfluous.

pmac:
My initial thoughts are that it's probably not going to help much with Basket. It's not really doing anything that'll help us. For a more complex app like Bedrock it could make things more clean and readable, but here it just seems to add complexity for no real gain. 

That said, if it's how you want to go for all of our Django stuff and it would be consistent with other projects then that's fine. I leave it to you and Steve to decide how you'd like to manage config.

stevejalim (commented):
Thanks for the experiment! I'm not averse to this, but I agree that in this case there's not a huge uplift to be had. (That said, I appreciate that this does make a neat, reasonably compact proof of concept too.)

re pmac's comment:
>  For a more complex app like Bedrock it could make things more clean and readable, but here it just seems to add complexity for no real gain.

I did try out django-configurations when doing the Pocket Mode stuff on Bedrock. I was hoping we could use inheritance to make it the settings easier to compose, but I remember bumping into some walls with it, the detail of which I can't recall, but may have been related to me trying to solve a problem while also grokking a new tool. I'd certainly welcome seeing if, with the experience you've got of d-c, you could make Bedrock run in two modes using it. 

Indeed, if we _can_ use d-c for Bedrock, I'd say it could be a lot easier/cleaner than the current approach with `settings/__init__.py` and imports, and - perhaps controversially - maybe we leave the other MEAO closer to stock Django + some env-var-getter?

robhudson:
I appreciate the feedback. Thanks y'all.

One thing I do like is the areas that got simplified, e.g.
```
SOME_SETTING = "this" if DEBUG else "that"
```
The above requires some mental processing, even if simple. Versus just reading the value set inside the appropriate settings class.

But those cases could potentially be cleaned up with a env var with a reasonable default.

Anyway, I just wanted to explore what this would look like. I'll go ahead and close this out.

PR #1262, opened 2024-01-04
Fix #1233: Move to uWSGI

### LINKED ISSUE #1233: Find a Replacement for Meinheld WSGI Server
## Description

Currently, this project relies on the [Meinheld](https://github.com/mopemope/meinheld) WSGI server as the underlying server technology, along with gunicorn, for serving our application. However, this project has not been updated, maintained, or seen a release in over 3 years. It also does not compile on Python versions >= 3.10.

robhudson:


bin/uwsgi.ini
```diff
@@ -0,0 +1,8 @@
+[uwsgi]
+module = basket.wsgi:application
+socket = :8001
```
stevejalim, line 3: How come port 8001 here? Does anything else need adjusting from 8000 to 8001 to match?

robhudson: Good call out. Reading the docs it's actually better to use a unix socket than a port. I might need some SRE guidance as to whether we have nginx in front of this or a load balancer to get these to line up. Oddly this worked fine when I deployed it to the dev server for testing.

bin/uwsgi.ini
```diff
@@ -0,0 +1,8 @@
+[uwsgi]
+module = basket.wsgi:application
+http = :8000
+master = true
+vacuum = true
+enable-threads = true
```
pmac, line 6: does this actually enable threads?

pmac: because we've had very bad results from enabling threading in Django apps in the past

robhudson: According to the docs:
```
By default the Python plugin does not initialize the GIL. This means your app-generated
threads will not run. If you need threads, remember to enable them with enable-threads.
```

So it is still in worker mode, but it allows Python to run threads in those workers.

pmac: but we don't want threads in the workers, do we? I just remember we tried running bedrock with several threads and workers and ended up with some very odd behavior, but that was many years ago with Apache mod_wsgi.

robhudson: We can run tests, but my hunch is we want this. There's enough going on with 3rd party libs and threads that I would be worried without it we'd run into trouble. Things like thread locals, and newrelic has thread profiling, and SQLite has thread sharing, etc. This isn't worker threads, from what I understand. It's actually disabling the GIL across the app if this isn't set to `true`.

PR #1263, opened 2024-01-06
Upgrade to Python 3.12.x and Django 4.2.x

robhudson:


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # BUILDER IMAGE
-FROM python:3.9-slim-bullseye AS builder
+FROM python:3.12-slim AS builder
```
stevejalim, line 2: Q Is there something in `3.12-slim` that isn't available in `-slim-bullseye`?

bin/uwsgi.ini
```diff
@@ -0,0 +1,8 @@
+[uwsgi]
+module = basket.wsgi:application
+http = :8000
+master = true
+workers = 8
```
stevejalim, line 5: If it helps, over in Nucleus I've left this out and set `--workers` in run-prod.sh so we can tune it with env vars. Might it be worth doing that here, in case we've got WSGI_NUM_WORKERS set to something special in k8s?

```uwsgi --ini /app/bin/uwsgi.ini --workers "${WSGI_NUM_WORKERS:-4}"```

**Review by stevejalim — APPROVED** [review]
Haven't test driven but code changes LGTM

r+wc

**Comment by stevejalim** [comment]
Just noticed when trying to run it all locally that it won't on my Cac:

```docker-compose run --rm test
image with reference redis was found but does not match the specified platform: wanted linux/amd64, actual: linux/arm64/v8
```

bin/uwsgi.ini
robhudson: I saw that. I will do the same 👍

Dockerfile
```diff
@@ -18,17 +18,18 @@ ENV DJANGO_SETTINGS_MODULE=basket.settings
 # Install app
 COPY requirements/* /app/requirements/
 
-# TODO: split out a separate dev image from the prod image and only install scoped deps
-# RUN pip install --require-hashes --no-cache-dir -r requirements/prod.txt
-RUN pip install --require-hashes --no-cache-dir -r requirements/dev.txt
+# The setuptools install is needed for pyfxa (currently v0.7.7) which calls `pkg_resources`,
+# and Python 3.12 no longer adds setuptools by default to the venv.
+RUN pip install -U setuptools && \
```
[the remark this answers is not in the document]
robhudson: setuptools is needed at "runtime" in basket b/c of pyfxa. pip-tools only installs it so it can compile the requirements but it doesn't get into the venv. So I still need this.

Dockerfile
robhudson: I updated to explicitly say `3.12-slim-bookworm` to also pin Debian releases.

stevejalim:
> Just noticed when trying to run it all locally that it won't on my Mac:


Update: now i've tweaked the Rosetta setting in my Docker, (see [here](https://github.com/getsentry/self-hosted/issues/1573#issuecomment-1838698365) it all works fine.

PR #1284, opened 2024-02-02
Add form clean for name field

robhudson:


basket/petition/forms.py
```diff
@@ -10,3 +12,9 @@ class PetitionForm(forms.ModelForm):
     class Meta:
         model = Petition
         fields = ["name", "email", "title", "affiliation"]
+
+    def clean_name(self):
+        name = self.cleaned_data["name"]
+        if re.search("<.*?>", name):
```
pmac, line 18: is this enough or should we use something like bleach?

robhudson: Bleach is no longer maintained, so we need to find an alternate solution for that still. So I'd prefer not to add it as a dependency at this point. I could do more here, however, e.g. only allow letters and common characters (`[.-_ ]`). Or, wait until we see more problems but I wanted to nail down this one quickly.

pmac (approved):

PR #1312, opened 2024-02-27
Add Braze client

robhudson:


basket/news/backends/braze.py
```diff
@@ -0,0 +1,181 @@
...
+        @raises: BrazeUnauthorizedError: 401 error (invalid API key)
+        @raises: BrazeForbiddenError: 403 error (invalid permissions)
+        @raises: BrazeNotFoundError: 404 error (invalid endpoint)
+        @raises: BrazeRateLimitError: 429 error (rate limit exceeded)
+        @raises: BrazeInternalServerError: 500 error (Braze server error)
+        @raises: BrazeClientError: any other error
+
+        """
+        if not self.active:
+            return
```
stevejalim, line 76: Question, non-blocking: do we want to log anything to flag this in logs?

basket/news/backends/braze.py
```diff
@@ -0,0 +1,181 @@
...
+
+        """
+        data = {
+            "user_aliases": [{"alias_name": email, "alias_label": "email"}],
+            "email_address": email,
+        }
+
+        return self._request(BrazeEndpoint.USERS_EXPORT_IDS, data)
+
+    def send_campaign(self, email, campaign_id):
```
stevejalim, line 158: Question, non-blocking: do we need a bulk version of this as well?

stevejalim (commented):
r+ wq

📤

stevejalim (approved):

basket/news/backends/braze.py
robhudson: There's a one-time warning on start up if the API key isn't supplied that should suffice. I just wanted it to be callable but no-op so the code integration can exist without throwing an error.

basket/news/backends/braze.py
robhudson: These are just transactional emails even though we are using the campaign API, so shouldn't be called in bulk. Kinda confusing, but apparently transactional emails are an add-on.

PR #1344, opened 2024-03-26
Add Braze integration

robhudson:


basket/news/tasks.py
```diff
@@ -471,12 +509,24 @@ def update_custom_unsub(token, reason):
 
 
 @rq_task
-def send_recovery_message_acoustic(email, token, lang, fmt):
-    message_name = "account-recovery"
-    if fmt != "H":
-        message_name += "-text"
+def send_recovery_message(email, token, lang):
```
pmac, line 512: we have to be careful changing task names because if we just deploy this then any tasks in the queue will be rejected since no such task name exists on new workers. It'd be better to keep the old name around that calls the new function for at least the initial deployment.

robhudson: Roger that. Keep scrolling down. ;)

PR #1350, opened 2024-04-03
Migrate to gunicorn + gevent

robhudson:


robhudson:
I'm seeing some Python 3.12 + gevent issues:

RuntimeError: can't create new thread at interpreter shutdown

Python 3.12 no longer allows creating a thread on interpreter shutdown:
https://github.com/python/cpython/pull/104826

stevejalim (approved):
r+ but why are we rolling this back from uwsgi?

robhudson:
Historical note: We switched to granian on bedrock so closing this for now until the dust settles.

PR #1352, opened 2024-04-04
Stop passing email format as it's not used in Braze

robhudson:


stevejalim:
Do these changes only affect the outbound post to Braze, or do they change what we capture from basket clients?

robhudson:
The format was never a required field, but this also removes the field from form processing of request to basket. It also removes passing it through to CTMS. Braze doesn't need it and never had it (since it's new). It doesn't ignore it from request to CTMS to get user data, so it will still live there but ignored for the most part.

stevejalim (approved):

PR #1387, opened 2024-05-10
Add email validation to fxa queue events

robhudson:


basket/news/tasks.py
```diff
@@ -106,14 +111,18 @@ def fxa_delete(data):
 def fxa_verified(data):
     """Add new FxA users"""
     # if we're not using the sandbox ignore testing domains
-    if email_is_testing(data["email"]):
+    email = process_email(data["email"])
+    if not email:
+        metrics.incr("news.tasks.invalid_fxa_email", tags=["task:fxa_verified"])
+        return
```
pmac, line 117: I'm not sure we should just throw away all information about these users. Are the email folks okay with doing this?

robhudson: The one you commented on here seems to only subscribe users to newsletters. But if the email is invalid then that subscription will do nothing. My assumption is we wouldn't want the invalid data.

pmac: Hmm. Is this not the main one that adds the user as an FxA user with their FxA ID?

pmac: I'm pretty sure this is indeed the one that records for marketing the fact that a user has created an FxA and what their ID is and all that.

robhudson: Okay, but how does it do that? This task just results in calling `upsert_contact(SUBSCRIBE, ...)`. There's also that `get_fxa_user_data` call and that will update the user's email, but do we want to be doing either of these things if the email is invalid? 

I do see that in `new_data` we have the fxa ID and that gets passed through in the upsert call. Should we perhaps drop the email field if it's invalid and send the rest instead?

pmac: > I do see that in `new_data` we have the fxa ID and that gets passed through in the upsert call. Should we perhaps drop the email field if it's invalid and send the rest instead?

Exactly. That bit creates the new contact and adds the FxA ID to the record which we expect to exist to look up the user's token if they click "email preferences" within the FxA UI itself. I think your question is the right question but we need to ask the email folks as I don't care either way, but they might, or they might decide that these FxAs with invalid emails are worthless and we don't need them at all. I'm very curious what FxA does with them.

PR #1393, opened 2024-05-17
Remove acoustic transactional code

robhudson:


basket/news/backends/acoustic.py
```diff
@@ -7,8 +7,6 @@
 from requests import ConnectionError
 from silverpop.api import Silverpop, SilverpopResponseException
 
-from basket import metrics
```
stevejalim, line 10: Where was this used? I don't see it in removed code so do we still need it?

stevejalim (approved):
r+ with a question

✉️

basket/news/backends/acoustic.py
robhudson: It was used down below on L127. The other code in this file don't use it so it can be removed.

stevejalim: d'oh serves me right for reviewing on GH Mobile

PR #1448, opened 2024-07-19
Fix #1446: Allow token-only subscribes

### LINKED ISSUE #1446: Allow subscriptions to API using basket token or email
Bedrock has a use-case where sending newsletter subscriptions with only a token would be desirable to avoid 2 steps - a user lookup by token, then a newsletter subscription by email.

robhudson:


pmac (approved):
Looks great! Thanks 💯

basket/news/tests/test_views.py
```diff
@@ -161,7 +172,7 @@ def test_empty_email_invalid(self, update_user_mock):
         resp = views.subscribe(req)
         resp_data = json.loads(resp.content)
         self.assertEqual(resp_data["status"], "error")
-        self.assertEqual(resp_data["code"], errors.BASKET_INVALID_EMAIL)
+        self.assertEqual(resp_data["code"], errors.BASKET_USAGE_ERROR)
```
robhudson, line 175: Slight change here since we're now checking for email or token and if either has a false-ish value we return the usage error. Previously if email was at least specified in the data but empty it would return invalid email. I consider this an acceptable change since email (or token) is required and an empty email is basically no email.

PR #1449, opened 2024-07-19
Fix #1447: Remove unused `subscribe_main` and `subscribe_json`

### LINKED ISSUE #1447: Remove `subscribe_json` and `subscribe_main`
Looking at our metrics, these views never get used and they add confusion on which subscribe view is the actual one being used. We should remove the dead code and the confusion.

robhudson:


pmac (approved):
🫗

PR #1467, opened 2024-08-01
Squash migrations

robhudson:
Django output says:

>   You should commit this migration but leave the old ones in place;
>   the new migration will be used for new installs. Once you are sure
>   all instances of the codebase have applied the migrations you squashed,
>   you can delete them.


stevejalim (approved):
This squashed migration is a lot smaller than I thought it would be for 32 migrations!

r+

pmac (approved):
Nice!

PR #1469, opened 2024-08-02
Fix #1466: Don't create a new user when primary email changes

### LINKED ISSUE #1466: Extra CTMS record created for FxA user
See [this diagram](https://gist.github.com/grahamalama/c4cb75fd6cc2e220e49ee630b3b6031b) for a flow chart of what is happening.

## What is currently happening

1. A user creates an FxA
2. User changes the primary email on their FxA
3. User clicks the link in the FxA UI to manage their email preferences
4. Basket gets that request and [attempts to look up the user by email address](https://github.com/mozmeao/basket/blob/fdede676203a90e92f85c0ee7845812ec8e7e8b4/basket/news/views.py#L169)
5. CTMS says there is no user with that email address because the user is in the system with their original email
6. Basket creates a new record in CTMS with the new email address 
7. Basket redirects the user to the email preferences page using the newly created CTMS entry data

## What should happen

For step 4 above, the first lookup that basket tries should be with the user's FxA ID. So we need to figure out if the FxA ID is part of the [data returned by FxA during the Oauth process](https://github.com/mozmeao/basket/blob/fdede676203a90e92f85c0ee7845812ec8e7e8b4/basket/news/views.py#L161).

robhudson:
When a user changes their primary email, without the `fxa_id` in the lookup, basket would query CTMS and get no response and treated this as a new user creating a new user in CTMS with the new email. By including the `fxa_id`, we are able to find the user and process the oauth flow correctly, sending the user on to the prefs center.

### DIFF
--- a/basket/news/tests/test_views.py
+++ b/basket/news/tests/test_views.py
@@ -888,7 +888,7 @@ class FxAPrefCenterOauthCallbackTests(ViewsPatcherMixin, TasksPatcherMixin, Test
         """Should return a redirect to email pref center"""
         fxa_oauth_mock, fxa_profile_mock = Mock(), Mock()
         fxa_oauth_mock.trade_code.return_value = {"access_token": "access-token"}
-        fxa_profile_mock.get_profile.return_value = {"email": "dude@example.com"}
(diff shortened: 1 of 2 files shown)


pmac (approved):
💯

PR #1471, opened 2024-08-28
Fix 1474: Delete petition app (all except migrations)

robhudson:
We need to keep migrations around to remove the tables. Once they are removed we can remove the rest of the app.


pmac (approved):
🗑️

PR #1482, opened 2024-09-04
Optimize the Dockerfile a bit

robhudson:
This attempts to reduce Docker layers by consolidating calls. E.g. multiple `ENV` calls are done in a single layer. Move `RUN` calls together as much as possible.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,58 +1,62 @@
 # BUILDER IMAGE
 FROM python:3.12-slim-bookworm AS builder
 
-# Extra python env
-ENV PYTHONDONTWRITEBYTECODE=1
-ENV PYTHONUNBUFFERED=1
-ENV PIP_DISABLE_PIP_VERSION_CHECK=1
-ENV PATH="/venv/bin:$PATH"
+# Set environment variables
+ENV PYTHONDONTWRITEBYTECODE=1 \
+    PYTHONUNBUFFERED=1 \
+    PIP_DISABLE_PIP_VERSION_CHECK=1 \
+    PATH="/venv/bin:$PATH" \
+    DJANGO_SETTINGS_MODULE=basket.settings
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,58 +1,61 @@
...
 COPY docker/bin/apt-install /usr/local/bin/
-RUN apt-install default-libmysqlclient-dev libxslt1.1 libxml2
+RUN <<EOT
+apt-install default-libmysqlclient-dev libxslt1.1 libxml2
+adduser --uid 1000 --disabled-password --gecos '' --no-create-home webdev
+chown webdev:webdev /app
+EOT
 
-ARG GIT_SHA=latest
-ENV GIT_SHA=${GIT_SHA}
```
robhudson, line 48: This 2 got moved up to the above `ENV` call with default.

Dockerfile
```diff
@@ -1,58 +1,61 @@
...
-RUN adduser --uid 1000 --disabled-password --gecos '' --no-create-home webdev
+# Switch to non-root user before copying files
+USER webdev
 
 COPY --from=builder /venv /venv
 COPY --from=builder /app /app
-
-# Change User
-RUN chown webdev.webdev -R .
-USER webdev
```
robhudson, line 58: Updated to switch to `webdev` user earlier which is good for security and also means the above `COPY`s will get set to the `webdev` user.

Dockerfile
pmac: But I think we still need the `ARG` line so that we can pass in the value during build

pmac: Yeah, it's passed in here:

https://github.com/mozmeao/basket/blob/5475580353a00dbe2ee26784aadb5fa3e96fa439/docker-compose.yml#L16

robhudson: Tested locally and added the `ARG` back in. Thanks!

PR #1510, opened 2024-10-11
Use uv in Dockerfile

robhudson:


pmac:
Could you add a description to this to tell us how this helps or changes things? For example, I'm not clear on how using `uv pip` instead of just `pip` helps or does things differently.

robhudson:
I was actually just experimenting and testing. I'll change to a draft PR but I'll end up closing this unless I find some reason to advocate for it.

PR #1515, opened 2024-10-18
Migrate from make to just

robhudson:
This can't be merged until we decide if we also want to update the github action to also use `just` or do something else with the CI commands. See `.github/workflows/pull_request_tests.yml`


stevejalim (approved):

Justfile
```diff
@@ -0,0 +1,85 @@
+DC := 'docker compose'
+DC_CI := 'bin/dc.sh'
```
pmac, line 2: I guess we don't need this if it's not going to have these commands

pmac (approved):
image

PR #1538, opened 2024-11-06
Cancel prev CI runs on new pushes

robhudson:
See:
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs

### DIFF
--- a/.github/workflows/pull_request_tests.yml
+++ b/.github/workflows/pull_request_tests.yml
@@ -5,6 +5,10 @@ on:
       - "main"
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   test-python:
     runs-on: ubuntu-latest


.github/workflows/pull_request_tests.yml
```diff
@@ -5,6 +5,10 @@ on:
       - "main"
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
pmac, line 9: This looks good. Can we also do this (or instead) with branch name? That way rapid pushes to e.g. `main` wouldn't kick off multiple builds/deploys.

robhudson: Yeah that's the `|| github.ref` part. If it's not a PR it'll use that.

stevejalim (approved):
R+

.github/workflows/pull_request_tests.yml
pmac: Ah! Nice!

PR #1541, opened 2024-11-08
Drop email format support

robhudson:
When unsent, this defaults to "H" in CTMS.


pmac (approved):
This looks fine. I believe it should be fine, but we should be sure that sending a call to subscribe that includes `format` does not cause an error, for compatibility with existing forms.

robhudson:
> we should be sure that sending a call to subscribe that includes `format` does not cause an error, for compatibility with existing forms.

Yeah. The existing API should but I'll make sure there's a test for it.

PR #1548, opened 2024-11-14
Fix rq exception handling for the failing state

robhudson:


basket/base/rq.py
```diff
@@ -174,11 +175,16 @@ def store_task_exception_handler(job, *exc_info):
...
         return
 
-    # A job will retry if it's failed but not yet reached the max retries.
-    # We know when a job is going to be retried if the status is `is_scheduled`, otherwise the
-    # status is set to `is_failed`.
+    # NOTE: We are deliberately not using `job.is_scheduled` or `job.is_failed` properties because
+    # they trigger a `get_status` call, which refreshes the status from Redis by default. Since the code
+    # is modifying the `job._status` property directly, Redis does not accurately reflect the job's
+    # status until processing completes. This custom exception handler is triggered in the middle of
+    # that process, so we must access the `_status` property directly to get the current state.
```
robhudson, line 182: @selwin - Is this something we might want to add to the documentation for custom exception handlers? This took me awhile to resolve. That, and the fact that `handle_exception` is called before `handle_job_failure` and the latter is what calls `job.set_status(JobStatus.FAILED, ...)`.

stevejalim (approved):
Great catch! Nice work

r+

PR #1554, opened 2024-11-22
Refactor custom exception handler

robhudson:
It was discovered that after fixing failure logging in the last commit, the sentry and metrics reporting around retries stopped working. Retries in RQ were still working but we were missing some of the logic around what we do in the event of a task failure that has retries left.

This refactors the logic with both retries and failures in mind and also tests for ignored errors and allowable retries.

Stepping through the logic we check for:
* Ignored errors: we log it to sentry and abort any remaining retries.
* Retries left but not allowed: log it to sentry, abort any remaining retries, store in FailedTask table.
* Retries left but allowed to retry: log it to sentry. RQ will retry.
* No retries left: This task has gone through all its retries, log to sentry and store in FailedTask table.

### DIFF
--- a/basket/base/rq.py
+++ b/basket/base/rq.py
@@ -6,17 +6,14 @@ from time import time
 from django.conf import settings
 
 import redis
-import requests
 import sentry_sdk
 from rq import Callback, Retry, SimpleWorker
 from rq.job import JobStatus
 from rq.queue import Queue
 from rq.serializers import JSONSerializer
-from silverpop.api import SilverpopResponseException
 
 from basket import metrics
 from basket.base.exceptions import RetryTask
-from basket.news.backends.common import NewsletterException
 
 # don't propagate and don't retry if these are the error messages
 IGNORE_ERROR_MSGS = [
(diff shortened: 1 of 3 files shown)


basket/base/rq.py
```diff
@@ -184,49 +183,59 @@ def store_task_exception_handler(job, *exc_info):
...
-        # Don't log to sentry if we explicitly raise `RetryTask`.
-        if not isinstance(exc_info[1], RetryTask):
-            with sentry_sdk.isolation_scope() as scope:
-                scope.set_tag("action", "retried")
-                sentry_sdk.capture_exception()
-
-    elif job._status == JobStatus.FAILED:
-        # Job failed but no retries left.
-        metrics.incr("base.tasks.failed", tags=[f"task:{task_name}"])
+    # Returning `False`` prevents any subsequent exception handlers from running.
```
stevejalim, line 218: **nitpick (non-blocking):** Would be good to disclose this at the top - I was wondering why earlier we returned `False`

stevejalim (approved):
r+wc

PR #1558, opened 2024-11-25
Bring order to the newsletter API

robhudson:
I noticed the ordering of the newsletter list wasn't ordered and traced it to the API missing this field on the ninja conversion.


pmac (approved):
📦

Issue #1530, opened 2024-10-31
EPIC: Migrate API to django-ninja


robhudson:
## Description:

Migrate basket API to django-ninja.

Django-ninja offers:
* **Built-in Data Validation**: Django Ninja offers automatic request and response validation using Pydantic
* **Type Safety**: With type hints, we can ensure type correctness on request and response data, making code more reliable and consistent across APIs and responses
* **Auto-Generated Documentation**: Django Ninja automatically generates an OpenAPI schema, making API documentation consistent and easy to update, and providing a user-friendly interface for using or testing the API

### Tasks

**Keep the same URL but move to django-ninja**
- [ ] `/news/newsletters/`
- [ ] `/news/subscribe/`: This is a form-encoded endpoint.
- [ ] `/news/unsubscribe/<uuid:token>/`
- [ ] `/news/custom_unsub_reason/`

**Move to new URL along with move to django-ninja**
- [ ] `/news/user/<uuid:token>/` -> `/users/<uuid:token>/`
- [ ] `/news/user-meta/<uuid:token>/` -> `/users/meta/<uuid:token>/`
- [ ] `/news/lookup-user/` -> `/users/lookup/`
- [ ] `/news/confirm/<uuid:token>/` -> `/users/confirm/<uuid:token>/` 
- [ ] `/news/recover/` -> `/users/recover/`

### TBD
- `/news/common-voice-goals/` - we still get data here.

All other URLs will remain as-is.

PR #1559, opened 2024-11-29
#1530: Add `/api/v1/users/recover/`

robhudson:
This also refactors the tests a bit as we add more APIs

### DIFF
--- a/basket/news/api.py
+++ b/basket/news/api.py
@@ -1,15 +1,26 @@
+import json
 import uuid
 
 from django.conf import settings
+from django.http import HttpResponse
 from django.views.decorators.cache import cache_page, never_cache
 
-from ninja import Router
+from ninja import NinjaAPI, Router
 from ninja.decorators import decorate_view
+from ninja.errors import ValidationError
 
 from basket import errors, metrics
+from basket.news import tasks
 from basket.news.auth import AUTHORIZED, FxaBearerToken, HeaderApiKey, QueryApiKey, Unauthorized
 from basket.news.models import Newsletter
-from basket.news.schemas import ErrorSchema, NewsletterSchema, NewslettersSchema, UserSchema
+from basket.news.schemas import (
+    ErrorSchema,
+    NewsletterModelSchema,
+    NewslettersSchema,
+    OkSchema,
+    RecoverUserSchema,
+    UserSchema,
+)
 from basket.news.utils import (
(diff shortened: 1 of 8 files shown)


basket/news/api.py
```diff
@@ -35,12 +39,50 @@
...
+    url_name="users.recover",
+    description="Send recovery email",
+    response={
+        200: OkSchema,
+        400: ErrorSchema,
+        404: ErrorSchema,
+        500: ErrorSchema,
+    },
+)
+def recover_user(request, email: Form[EmailStr]):
```
robhudson, line 62: I'm debating moving to a JSON only API. Bedrock should be able to migrate to that easily enough but I'm curious how far this is used outside of Bedrock. It may mean keeping the old path around.

pmac: I wouldn't think this is used outside of bedrock. JSON only +1 from me.

basket/urls.py
```diff
@@ -50,3 +55,20 @@
             path("500/", defaults.server_error),
         ]
     )
+
+
+# So django-ninja returns a pydantic validation error in a consistent JSON shape.
+@api.exception_handler(ValidationError)
```
pmac, line 61: would this make more sense to be in the `api.py` file, or does it need to be in `urls.py` for some reason?

basket/news/api.py
stevejalim: Same here - JSON all the way

basket/news/tests/api/test_newsletters.py
```diff
@@ -12,11 +12,13 @@
 @pytest.mark.django_db
 class TestNewslettersAPI:
     def setup_method(self, method):
-        cache.clear()
         self.url = reverse("api.v1:news.newsletters")
         self.n1 = self._add_newsletter("test-1", show=True, order=1)
         self.n2 = self._add_newsletter("test-2", order=2)
 
+    def teardown_method(self, method):
+        cache.clear()
```
stevejalim, line 20: Worth keeping the cache.clear() in the setup too, in case another test poisons it?

stevejalim (approved):
r+

basket/urls.py
robhudson: Yeah that makes sense to me. The `api` object is created here but could probably be imported there or created there and imported here. I'll play with it.

PR #1567, opened 2024-12-02
chore(workflow): add build and push workflow

Basma1912:


.github/workflows/build-and-push.yml
```diff
@@ -0,0 +1,132 @@
...
+  GAR_LOCATION: us
+  GCP_PROJECT_ID: moz-fx-basket-prod
+  GAR_REPOSITORY: basket
+  REF_ID: ${{ github.ref }}
+
+
+jobs:
+  build_and_publish_public_images:
+    name: Build public basket images and push to Docker hub
+    runs-on: [self-hosted, prod, us-central1, bedrock]
```
robhudson, line 29: Is the `bedrock` here correct?

robhudson:
To share what I asked about in Slack:

I'm concerned about the subtle difference between the main and prod branches for the same file. I believe the `git push HEAD:prod` that the `bin/tag-release.sh` file executes will overwrite this file with the stage/dev workflow. 

Also, us devs never work on the prod branch. It's used as a tag and snapshot to trigger prod pushes and we consider it identical to the tagged release.

If there's a different way to set this up I'm open to moving to a deploy that is consistent between bedrock, basket, and ctms since the same group of people work on all three. Consistency would be a nice end-goal since sometimes it is difficult to mentally keep track of the differences.

robhudson:
prod branch not needed since the deploy pushes to the prod branch

PR #1577, opened 2024-12-11
Migrate to everett

robhudson:
This is for consistency across meao projects (bedrock, et al)

### DIFF
--- a/basket/settings.py
+++ b/basket/settings.py
@@ -10,12 +10,18 @@ import dj_database_url
 import django_cache_url
 import markus
 import sentry_sdk
-from decouple import Csv, UndefinedValueError, config
+from everett.manager import ChoiceOf, ConfigManager, ConfigurationMissingError, ListOf
 from sentry_processor import DesensitizationProcessor
 from sentry_sdk.integrations.django import DjangoIntegration
 from sentry_sdk.integrations.logging import ignore_logger
 from sentry_sdk.integrations.redis import RedisIntegration
 from sentry_sdk.integrations.rq import RqIntegration
+
+# The `basic_config` manager searches in this order:
+#   1. environment variables
+#   2. .env file
+#   3. `default` keyword argument
+config = ConfigManager.basic_config()
 
 # Application version.
 VERSION = (0, 1)
@@ -29,9 +35,9 @@ def path(*args):
     return str(ROOT_PATH.joinpath(*args))
 
 
-LOCAL_DEV = config("LOCAL_DEV", False, cast=bool)
-DEBUG = config("DEBUG", default=False, cast=bool)
-UNITTEST = config("UNITTEST", default=False, cast=bool)
+LOCAL_DEV = config("LOCAL_DEV", parser=bool, default="false")
+DEBUG = config("DEBUG", parser=bool, default="false")
+UNITTEST = config("UNITTEST", parser=bool, default="false")
 
 # If we forget to set a `UNITTEST` env var by are running `pytest`, set it.
 if sys.argv[0].endswith(("py.test", "pytest")):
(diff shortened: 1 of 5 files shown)


basket/settings.py
```diff
@@ -85,26 +88,27 @@ def path(*args):
...
 EMAIL_SUBJECT_PREFIX = config("EMAIL_SUBJECT_PREFIX", default="[basket] ")
 EMAIL_HOST_USER = config("EMAIL_HOST_USER", default="")
 EMAIL_HOST_PASSWORD = config("EMAIL_HOST_PASSWORD", default="")
 
 ALLOWED_HOSTS = config(
     "ALLOWED_HOSTS",
-    default=".allizom.org, .moz.works, basket.mozmar.org, basket.mozilla.com, basket.mozilla.org",
-    cast=Csv(),
+    parser=ListOf(str, allow_empty=False),
+    default="basket-dev.allizom.org,basket.allizom.org,basket.mozilla.org",
```
stevejalim, line 100: Just calling out that these defaults are different, but I can see you've updated them to current vals

basket/settings.py
```diff
@@ -85,26 +88,27 @@ def path(*args):
...
+)
+ALLOWED_CIDR_NETS = config(
+    "ALLOWED_CIDR_NETS",
```
stevejalim, line 103: In the switch from Csv() to ListOf(str), above, you take out the spaces in the comma-separated list. Do we need to also update `ALLOWED_CIDR_NETS` in the relevant env vars to not have spaces, too?

basket/settings.py
```diff
@@ -396,54 +392,48 @@ def before_send(event, hint):
...
 
 if UNITTEST:
     TESTING_EMAIL_DOMAINS = []
 else:
     TESTING_EMAIL_DOMAINS = config(
         "TESTING_EMAIL_DOMAINS",
-        "restmail.net,restmail.lcip.org,example.com",
-        cast=Csv(),
+        parser=ListOf(str, allow_empty=False),
+        default="restmail.net,restmail.lcip.org,example.com",
```
stevejalim, line 407: What's `restmail.lcip.org`?

requirements/dev.txt
```diff
@@ -300,15 +300,13 @@ cryptography==44.0.0 \
     --hash=sha256:3c672a53c0fb4725a29c303be906d3c1fa99c32f58abe008a82705f9ee96f40b \
     --hash=sha256:404fdc66ee5f83a1388be54300ae978b2efd538018de18556dde92575e05defc \
     --hash=sha256:4ac4c9f37eba52cb6fbeaf5b59c152ea976726b865bd4cf87883a7e7006cc543 \
-    --hash=sha256:60eb32934076fa07e4316b7b2742fa52cbb190b42c2df2863dbc4230a0a9b385 \
```
stevejalim, line 303: Any idea why these hashes have come out for `cryptography`? Should they be in this changeset?

stevejalim (approved):
LGTM

r+wc

basket/settings.py
pmac: It was a restmail instance, and it looks like Mozilla still owns `lcip.org`, but nothing is there anymore. May be outdated.

basket/settings.py
robhudson: Yeah we had some old stuff in there

basket/settings.py
robhudson: We don't seem to set this env var in any of the environments. But I just removed the spaces b/c it's a CSV string and it matched my mental model of what that should look like. But everett removes the spaces anyway:

```python
>>> ListOf(str)('foo, bar,  baz')
['foo', 'bar', 'baz']
```

basket/settings.py
robhudson: Probably best to remove it if nothing lives there.

requirements/dev.txt
robhudson: It's a re-compile of the requirements so I assumed some build changed, but I can remove them from this PR since it's unrelated.

PR #1593, opened 2024-12-17
Migrate to granian

robhudson:
This also drops newrelic since newrelic doesn't support granian.


bin/run-dev.sh
```diff
@@ -2,4 +2,14 @@
...
+granian --interface wsgi \
+    --host 0.0.0.0 \
+    --port "${PORT:-8000}" \
+    --no-ws \
+    --workers "1" \
+    --threads "1" \
+    --log-level "${GRANIAN_LOG_LEVEL:-debug}" \
+    --access-log \
+    --reload \
+    basket.wsgi:application
```
robhudson, line 15: I could be convinced not to go this route, but I do like it to flex granian while developing. I tested the reload and it works well -- even for template changes.

robhudson:
This should be merged first to set up the Granian config env vars:
https://github.com/mozilla-it/webservices-infra/pull/3614

stevejalim (approved):
LGTM

PR #1596, opened 2024-12-18
Update python dependencies

robhudson:
This also updates the checker so something like `apscheduler` from the `uv` output matches the `APScheduler` from the requirements file.


pmac (approved):
📦

PR #1602, opened 2024-12-24
Add DSAR admin page

robhudson:
This adds a DSAR page to the admin.

Step 1: Empty page
dsar-1

Step 2: Enter emails
dsar-2

Step 3: Output (with copy button)
dsar-3

### DIFF
--- a/basket/admin.py
+++ b/basket/admin.py
@@ -1,0 +1,6 @@
+from django.contrib import admin
+
+
+class BasketAdminSite(admin.AdminSite):
+    site_title = site_header = "Basket"
+    index_template = "admin/basket_index.html"
--- a/basket/base/forms.py
+++ b/basket/base/forms.py
@@ -9,12 +9,7 @@ class EmailListField(forms.Field):
         if not value:
             return []
 
-        emails = []
-        for email in [v.strip() for v in value.split("\n")]:
-            if email:
-                emails.append(email)
-
-        return emails
+        return [email.strip() for email in value.splitlines() if email.strip()]
 
     def validate(self, value):
         """Check if value consists only of valid emails."""
--- a/basket/base/forms.py
+++ b/basket/base/forms.py
@@ -1,0 +1,37 @@
+from django import forms
+from django.core.exceptions import ValidationError
+from django.core.validators import validate_email
+
+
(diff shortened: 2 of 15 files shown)


basket/base/views.py
```diff
@@ -0,0 +1,49 @@
...
+from django.shortcuts import render
+
+from basket.base.forms import EmailListForm
+from basket.news.backends.ctms import (
+    CTMSNotFoundByEmailError,
+    ctms,
+)
+
+
+@staff_member_required
```
robhudson, line 11: I'm thinking it might be nice to also make this require certain permissions so not everyone with a Basket login sees this.

robhudson: Added the DSAR user permission and group to the 2nd commit on this PR.

basket/admin.py
```diff
@@ -0,0 +1,6 @@
+from django.contrib import admin
```
pmac, line 1: Looks like this file and the `apps.py` work here, but I thought they went in app folders like `base` where the rest of the stuff in the PR went. No?

basket/base/forms.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class EmailListField(forms.Field):
+    def to_python(self, value):
+        """Normalize data to a list of strings."""
+        if not value:
+            return []
+
+        emails = []
+        for email in [v.strip() for v in value.split("\n")]:
```
pmac, line 13: I think it'd be safer to use `value.splitlines()`, though using `split("\n")` and `strip()` is probably roughly equivalent.

pmac (approved):
Just a couple comments you can take or leave. Thanks for knocking this out!

basket/admin.py
robhudson: It's here b/c I defined it to be here based on the `basket/apps.py` file pointing to this. It seemed like a good general place to put it that isn't app specific.

basket/base/forms.py
robhudson: It doesn't `strip()` but it probably supports other line boundaries, so I'll switch to `splitlines()`. 👍

robhudson:
Updated to add forms and view tests, and review fixes.

PR #1603, opened 2024-12-27
Use "testqueue" for unit tests

robhudson:
I noticed that when I run basket locally, it often had tasks being retried from the test runs. The retried tasks were being left scheduled in Redis after a test run, so running basket would retry them.

This updates the default queue in tests to use a separate queue. Apparently the `queue.empty()` only clears the active queue, not the `ScheduledJobRegistry`. It looks to be a bit of a pain to remove scheduled jobs, but not impossible. I can add this at some future time.

Regardless, it's probably a better idea to use a separate queue for testing just like we have a separate db for testing.

### DIFF
--- a/basket/base/rq.py
+++ b/basket/base/rq.py
@@ -55,11 +55,19 @@ def get_redis_connection(url=None, force=False):
     return _REDIS_CONN
 
 
-def get_queue(queue="default"):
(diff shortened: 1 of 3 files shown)


stevejalim (approved):
LGTM!

PR #1608, opened 2024-12-31
#1530: Add /api/v1/users/confirm/

robhudson:
Issue #1530 

This adds `/api/v1/users/confirm/` and refactors a few things a bit in the process.


stevejalim (approved):
r+

PR #1623, opened 2025-01-15
Fix #1620 - Add `last_accessed` column to `APIUser`

### LINKED ISSUE #1620: Add `last_access` timestamp to `APIUser` model and log access
Introduce a `last_access` field to the `APIUser` model and update it whenever an API user makes a request. The purpose of this change is to track the last time an API user accessed the system, which can be useful for auditing, troubleshooting, and deactivating stale accounts.

A `created` date might be nice while we're add it.

robhudson:


basket/news/models.py
```diff
@@ -140,6 +141,8 @@ class APIUser(models.Model):
     name = models.CharField(max_length=256, help_text="Descriptive name of this user")
     api_key = models.CharField(max_length=40, default=get_uuid, db_index=True)
     enabled = models.BooleanField(default=True)
+    created = models.DateTimeField(auto_now_add=True)
```
pmac, line 144: I remember it being the generally accepted advice to avoid using `auto_now_add` in favor of something like `default=now` where `now` was a reference to the function in datetime. Is that no longer the case?

pmac (approved):
Looks great. Just had that one question about `auto_now_add`.

basket/news/models.py
robhudson: Yep, I remember that but for this type of field I didn't see any issues with it. It gets set once on creation. Under the hood Django also makes the field non-editable. And you can't create an instance with your own `created` datetime but I didn't see either of those as issues in this case.

PR #1636, opened 2025-02-04
Add DSAR unsubscribe users and fetch user info admin pages

pmac:
I've refactored how the new admin pages are done. By adding them via the custom `AdminSite` class we get the pages listed in the same way as the other admin pages. We also get the full admin context, which gives us the sidebar on the custom pages.


pmac:
Screenshot of me testing locally
image

basket/base/templates/admin/dsar-info.html
```diff
@@ -0,0 +1,183 @@
...
+      {% if dsar_contact %}
+        <h2>User Info:</h2>
+        <table>
+            <tr>
+                <th>Primary Email</th>
+                <td>{{ dsar_contact.email.primary_email }}</td>
+            </tr>
+            <tr>
+                <th>Basket Token</th>
+                <td><a href="https://www.mozilla.org/newsletter/existing/{{ dsar_contact.email.basket_token }}/" target="_blank" title="Open the user's Email Preferences">{{ dsar_contact.email.basket_token }}</a></td>
```
robhudson, line 80: Might be nice to add a constant for the pref center path. Not env var based but just make it easy to find if we ever move it.

basket/base/templates/admin/dsar-info.html
```diff
@@ -0,0 +1,183 @@
...
+                    {% for nl in newsletter_names %}
+                        <li>{{ nl }}</li>
+                    {% empty %}
+                        <li>No subscriptions</li>
+                    {% endfor %}
+                    </ul>
+                </td>
+            </tr>
+        </table>
+        <br>
```
robhudson, line 133: All of the below content looks to be duplicated from the dsar.html template. Perhaps we could pull it out into a template include and then pull it into each template to reduce duplication.

robhudson (approved):
Great stuff. Looks like good useful tools. I tested locally and everything works well. Just a couple little suggestions.

pmac:
Good ideas. Thanks for taking a look.

pmac:
Updated with your suggestions

PR #1639, opened 2025-03-26
Update Django to 5.0.x

robhudson:
On the way to 5.2


pmac (approved):
Much more straight-forward than I expected. Nice!

PR #1643, opened 2025-04-15
Update pre-commit config

robhudson:
Adding some common pre-commit checks, plus Python and Django upgrade.


.git-blame-ignore-revs
```diff
@@ -1,3 +1,2 @@
 # Migrate code style to Black
 de6a33a03040e370fc5701622cb2496c7ece4cff
```
pmac, line 2: Why remove this?

robhudson: It's there it's just fixing up the end of file. Weird that it's formatted that way though.

pmac: Huh. Weird. Github made it look like the hash was removed.

pmac (approved):

PR #1694, opened 2026-03-13
fix(redis): fixes a race condition where jobs cant be found due to 0 ttl

stephendherrera:
Fix RQ worker cleanup errors that surfaced today due to unknown cause (potentially slower external service response times) exposing a latent bug where result_ttl=0 causes job keys to be deleted before cleanup completes. Changed default to 60 seconds.


stevejalim (approved):
Makes sense to me. r+ 

Any idea why we had `Ignore results.` originally?

PR #1716, opened 2026-04-10
Check for missing email down certain code paths

bluewave41:
Sentry is throwing errors for data["email"] not existing when trying to send tx messages. This PR adds email fetching if that path is entered just in case.


stephendherrera (approved):
Approved to test in stage.

PR #1760, opened 2026-06-08
feat(braze): add endpoint for webhook to assign external_ids when mis…

stephendherrera:
Adds `POST /api/v1/users/assign/`, an authenticated webhook (bearer secret) that Braze calls for users who have only a user alias and no external_id. Basket assigns one reusing the basket_token when it's a valid UUID, otherwise minting a fresh UUID and writes it back via Braze /users/identify. fxa-only/email-only users also get a basket_token alias equal to their external_id, added via a delayed task so Braze can propagate the new id first. Backend-gated by the BRAZE_*_WRITE flags; non-retryable Braze errors are logged + metriced instead of failing the job. (WT-1280)

### DIFF
--- a/basket/base/tests/test_throttling.py
+++ b/basket/base/tests/test_throttling.py
@@ -1,6 +1,8 @@
+from types import SimpleNamespace
+
 import pytest
 
-from basket.base.throttling import MultiPeriodThrottle
+from basket.base.throttling import MultiPeriodThrottle, WebhookIdentifierThrottle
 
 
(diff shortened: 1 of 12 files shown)


stevejalim (approved):
LGTM

I wonder if there's any risk of identify_user chewing through rate limits - if we get a ton of users that need identifying does that basically double our Braze API calls (maybe not?) - and if we double them, does our webhook rate limit already have that accounted for?

r+withthecommentabove :)

PR #1762, opened 2026-06-12
feat(contact): adjust enterprise form fields and tests

stephendherrera:


bluewave41 (approved):

basket/contact/api.py
```diff
@@ -39,9 +39,9 @@ def contact_enterprise(request, payload: ContactEnterpriseSchema):
         metrics.incr("contact.enterprise.ratelimited", tags=["reason:email"])
         return 429, {"status": "error"}
 
-    if payload.website:
+    if payload.office_fax:
         metrics.incr("contact.enterprise.honeypot")
         return {"status": "ok"}
 
-    tasks.submit_contact.delay(payload.model_dump(exclude={"website"}))
+    tasks.submit_contact.delay(payload.model_dump(exclude={"website", "office_fax"}))
```
stevejalim, line 46: Should re retain `website` in this file or just replace it with `office_fax` - I don't have context to know, but mentioning it just in case

stevejalim (approved):
r+wc

basket/contact/api.py
stephendherrera: Thanks @stevejalim. Keeping for now should we need to include in the form submission in coming weeks.

PR #1766, opened 2026-08-11
Remove cTms fallthrough paths

bluewave41:
This PR removes cTms code branches to make Braze the sole backend provider.

### DIFF
--- a/basket/news/backends/braze.py
+++ b/basket/news/backends/braze.py
@@ -445,6 +445,9 @@ class Braze:
 
         @param data: user data to add as a new user.
         """
+        if not self.interface.active:
+            raise BrazeNotConfigured()
+
         braze_user_data = self.to_vendor(None, data)
         external_id = braze_user_data["attributes"][0]["external_id"]
         self.interface.save_user(braze_user_data)
@@ -474,6 +477,9 @@ class Braze:
         @param existing_data: current user record
         @param update_data: dict of new data
         """
+        if not self.interface.active:
+            raise BrazeNotConfigured()
+
         braze_user_data = self.to_vendor(existing_data, update_data)
         external_id = braze_user_data["attributes"][0]["external_id"]
         self.interface.save_user(braze_user_data)
@@ -519,6 +525,9 @@ class Braze:
         @return: deleted user data if successful
         @raises: BrazeUserNotFoundByEmailError
         """
+        if not self.interface.active:
+            raise BrazeNotConfigured()
+
         data = self.interface.export_users(email=email, fields_to_export=["external_id", "user_aliases"])
         if not data["users"]:
             raise BrazeUserNotFoundByEmailError
--- a/basket/news/backends/braze.py
+++ b/basket/news/backends/braze.py
@@ -391,6 +391,8 @@ class Braze:
         @param fxa_id: external ID from FxA
         @return: dict, or None if not found
         """
+        if not self.interface.active:
+            raise BrazeNotConfigured()
 
         user_response = self.interface.export_users(
             email,
--- a/basket/news/backends/braze.py
+++ b/basket/news/backends/braze.py
@@ -14,8 +14,8 @@ from tenacity import retry, retry_if_exception_type, stop_after_attempt, wait_fi
 from basket import metrics
 from basket.base.decorators import rq_task
 from basket.base.utils import is_valid_uuid
-from basket.news.backends.ctms import ctms, process_country, process_lang
-from basket.news.newsletters import newsletter_obj, slug_to_vendor_id, vendor_id_to_slug
+from basket.news.country_codes import SFDC_COUNTRIES_LIST, convert_country_3_to_2
+from basket.news.newsletters import is_supported_newsletter_language, newsletter_obj, slug_to_vendor_id, vendor_id_to_slug
 
 log = logging.getLogger(__name__)
 
@@ -91,6 +91,13 @@ class BrazeEndpoint(Enum):
     SUBSCRIPTION_USER_STATUS = "/subscription/user/status"
     USERS_ADD_ALIAS = "/users/alias/new"
     USERS_IDENTIFY = "/users/identify"
+
+
+class BrazeNotConfigured(Exception):
+    """Braze is not configured."""
+
+    def __str__(self):
+        return "Braze is not configured"
 
 
 class BrazeInterface:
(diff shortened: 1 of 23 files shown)


basket/settings.py
```diff
@@ -66,6 +66,7 @@ def path(*args):
 
 # CACHE_URL and RQ_URL are derived from REDIS_URL.
 REDIS_URL = config("REDIS_URL", default="")
+RQ_URL = None
```
bluewave41: remove this

basket/news/views.py
```diff
@@ -896,38 +689,12 @@ def lookup_user(request):
...
-                use_braze_backend=True,
-            )
-        else:
-            user_data = get_user_data(
-                token=token,
-                email=email,
-                masked=not authorized,
-                use_braze_backend=False,
-            )
+        user_data = get_user_data(
```
bluewave41, line 692: This one I'm really not sure about. I did this for one of the tests but I need to revisit the omit extra fields here cause I don't think this is right

basket/news/views.py
```diff
@@ -896,38 +689,12 @@ def lookup_user(request):
...
+            token=token,
+            email=email,
+            masked=not authorized,
+            omit_extra_braze_fields=True,
```
stephendherrera: I think this should be `not authorized`, like both branches this replaced. `True` looks to make utils.py:324 fall back to `ALLOWED_USER_FIELDS`, so API-key callers lose `email_id`, `fxa_id`, `unsub_reason`, and the rest.

basket/news/tasks.py
```diff
@@ -606,10 +526,10 @@ def record_common_voice_update(data):
         new_data["cv_" + k] = v
 
     if user_data:
-        ctms.update(user_data, new_data)
+        braze.update(user_data, new_data)
```
stephendherrera, line 529: This won't subscribe anyone. Line 523 passes newsletters as a list, and Braze's converter only handles them when they're a dict and silently skips them otherwise. CTMS used to convert a list to a dict for you, so this breaks in the switch to braze. Pass `{settings.COMMON_VOICE_NEWSLETTER: True}` instead. The tests mock out `braze` entirely, so they don't catch it.

basket/news/backends/ctms.py
```diff
@@ -980,15 +980,15 @@ def delete(self, email):
...
-    if settings.CTMS_ENABLED:
-        return CTMSSession(
-            api_url=settings.CTMS_URL,
-            client_id=settings.CTMS_CLIENT_ID,
-            client_secret=settings.CTMS_CLIENT_SECRET,
-        )
-    else:
-        logger.warning("CTMS not enabled.")
-        return None
+    # if settings.CTMS_ENABLED:
```
stephendherrera: Can we just delete? This now returns `None` implicitly and `test_ctms_session_disabled` below asserts that.

basket/news/tests/test_ctms.py
```diff
@@ -1051,7 +1051,7 @@ def test_init_long_api_url(self):
         )
         assert session.api_url == "https://ctms.example.com"
 
-    @override_settings(
+    '''@override_settings(
```
stephendherrera: Can we delete this?

basket/news/backends/braze.py
```diff
@@ -93,6 +93,13 @@ class BrazeEndpoint(Enum):
     USERS_IDENTIFY = "/users/identify"
 
 
+class BrazeNotConfigured(Exception):
```
stephendherrera, line 96: I don't think anything raises this exception. We can raise when `BRAZE_API_KEY` is empty, or drop the class and the `except`.

basket/settings.py
stephendherrera: Can we put a comment here on why we did this?

stephendherrera (changes requested):
Good work @bluewave41.

Two items I'd consider blocking: views.py:696 and tasks.py:529.

I imagine we'll leave this on stage for a while and observe? Related: this drops the CTMS shim in `braze.get`, so a `token`/`fxa_id` lookup for a user whose Braze aliases were never migrated now 404s with no fallback. Is that migration finished?

Also wondering if we can clean up some of the CTMS naming that's left in the kept code. The `get_user_data` docstring at utils.py:253 still says CTMS is the primary source, and test_users.py and tests/api/__init__.py kept `ctms_mock` arg names and `test_ctms_*` method names while switching the patch target to braze.

basket/settings.py
bluewave41: That's a note for me. I added it and we don't actually need it.

basket/news/views.py
bluewave41: Yep I left a note for myself to come back to this. The fxa_id was missing and the omission skipped it to pass but masked the actual problem.

basket/news/tasks.py
stephendherrera: Re-flagging: I know commonvoice is largely sunset, but if we can keep this updated in case they bring it back that would be preferred. this one is still here.

basket/news/backends/ctms.py
```diff
@@ -980,15 +980,6 @@ def delete(self, email):
 
 def ctms_session():
     """Return a CTMSSession configured from Django settings."""
```
stephendherrera, line 982: Function body is now just a docstring. Can we `return None` with a comment, or drop `ctms_session`/`ctms_interface` entirely.

basket/news/tests/test_ctms.py
```diff
@@ -1051,17 +1051,6 @@ def test_init_long_api_url(self):
         )
         assert session.api_url == "https://ctms.example.com"
```
stephendherrera, line 1052: This overrides `CTMS_ENABLED`, which this PR deleted, to assert that a function that can only return `None` returns `None`. Can go.

basket/news/backends/braze.py
```diff
@@ -384,6 +391,8 @@ def get(
         @param fxa_id: external ID from FxA
         @return: dict, or None if not found
         """
+        if not self.interface.active:
+            raise BrazeNotConfigured()
```
stephendherrera, line 396: Good catch. 

`add`/`update`/`delete` don't have the same guard, so writes still silently no-op while reads now fail loudly. And nothing covers this since the `braze_client` fixture always has a key; `BrazeInterface("http://test.com", "")` plus an assert that `get()` raises would lock it in.

stephendherrera (changes requested):
I'd fix the tasks.py:529 if possible. Otherwise, left a few nits, none of them blocking.

stephendherrera (approved):
Ok for stage testing, but one thing i'd still like to know before this goes past stage, with the CTMS shim gone from braze.get, a token or fxa_id lookup for someone whose braze aliases never got migrated will 404 with no fallback. is that migration actually done and I'm just missing?

PR #1768, opened 2026-09-02
Add basket_token alias to lookup

bluewave41:
Add basket_token alias to user lookups so we can query by basket_token.

### DIFF
--- a/basket/news/backends/braze.py
+++ b/basket/news/backends/braze.py
@@ -250,13 +250,16 @@ class BrazeInterface:
         if external_id:
             data["external_ids"] = [external_id]
             data["user_aliases"].append({"alias_name": external_id, "alias_label": "basket_token"})
-            data["user_aliases"].append({"alias_name": external_id.replace("-", ""), "alias_label": "fxa_id"})
 
         if fields_to_export:
             data["fields_to_export"] = fields_to_export
 
         if fxa_id:
             data["user_aliases"].append({"alias_name": fxa_id, "alias_label": "fxa_id"})
+        elif external_id:
+            # if we're trying to look up a user based on an fxa_id a hyphenated UUID is given
+            # Braze stores unhyphenated UUIDs so we need to handle that here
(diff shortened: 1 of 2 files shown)


basket/news/backends/braze.py
```diff
@@ -249,6 +249,8 @@ def export_users(self, email, fields_to_export=None, external_id=None, fxa_id=No
 
         if external_id:
             data["external_ids"] = [external_id]
+            data["user_aliases"].append({"alias_name": external_id, "alias_label": "basket_token"})
+            data["user_aliases"].append({"alias_name": external_id.replace("-", ""), "alias_label": "fxa_id"})
```
stephendherrera, line 253: Is this meant only as a fallback for the case where the URL "token" is actually a dash-less fxa_id (tying into the new UUID converter)? If so, can we gate it to only fire when fxa_id isn't already provided, or add a comment explaining the intent? I actually thought this might be a bug, but perhaps I'm just confused.

basket/news/urls.py
```diff
@@ -14,13 +16,26 @@
...
 
+
+class UUIDConverter:
+    regex = r"[0-9a-fA-F]{8}-?[0-9a-fA-F]{4}-?[0-9a-fA-F]{4}-?[0-9a-fA-F]{4}-?[0-9a-fA-F]{12}"
+
+    def to_python(self, value):
+        return uuid.UUID(value)
+
+    def to_url(self, value):
+        return str(uuid.UUID(str(value)))
```
stephendherrera, line 27: Not a blocker. `str(uuid.UUID(str(value)))` goes through this even if the value is a UUID. Could simplify if you'd like.

stephendherrera (changes requested):

basket/news/backends/braze.py
bluewave41: That is the intent. The fxa_id in Braze isn't hyphenated but the converter converts it to use hyphens to normalize it so I don't break anything else that might expect it to be a UUID later.

stephendherrera (approved):
r+

PR #1771, opened 2026-09-14
chore(log): config updates

stephendherrera:


stevejalim (approved):
r+ but wondering why we need the special UUID conversion

PR #1773, opened 2026-09-15
Wt 959 remove double path

bluewave41:


basket/news/models.py
```diff
@@ -221,8 +223,18 @@ def formatted_call(self):
...
     def retry(self):
+        task_kwargs = self.kwargs
+
+        # fxa_* tasks accept a catch-all **kwargs for extra data we don't want
+        # to replay; drop anything not in the task's named parameters.
+        if self.name.rsplit(".", 1)[-1].startswith("fxa_"):
+            func = import_string(self.name)
+            params = inspect.signature(func).parameters
+            known_params = {name for name, p in params.items() if p.kind != inspect.Parameter.VAR_KEYWORD}
+            task_kwargs = {key: val for key, val in task_kwargs.items() if key in known_params}
```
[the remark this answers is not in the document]
stephendherrera: @bluewave41 I think copilot is correct here. Looks to apply to QueuedTask.retry(). Don't see any kwarg filtering, not even for fxa_*. Any pre-rollout task with use_braze_backend in its kwargs (confirm_user, upsert_user, update_custom_unsub, update_user_meta) will TypeError on retry there. Should filter both methods, not just FailedTask.

basket/news/utils.py
```diff
@@ -300,20 +293,17 @@ def get_user_data(
...
-                fxa_id=fxa_id,
-            )
-    except CTMSNotFoundByAltIDError:
-        return None
+        backend_user = braze.get(
+            token=token,
+            email=email,
+            fxa_id=fxa_id,
+        )
+    except BrazeNotConfigured as exc:
```
stephendherrera: I'd suggest replacing this with handling for those Braze exception types (mapped to NewsletterException, similar to the BrazeNotConfigured case), otherwise real Braze failures propagate unhandled.

basket/news/tests/test_upsert_user.py
```diff
@@ -603,11 +603,11 @@ def test_existing_mofo_user_subscribes_to_mofo_newsletter(
         update_data = data.copy()
         update_data["newsletters"] = {"mozilla-foundation": True}
         update_data["optin"] = True
-        ctms_mock.update.assert_called_with(user_data, update_data)
+        braze_mock.update.assert_called_with(user_data, update_data)
         confirm_mock.delay.assert_not_called()
 
     @patch("basket.news.tasks.send_tx_messages")
-    def test_send_transactional(self, braze_mock, get_user_mock, ctms_mock, confirm_mock):
+    def test_send_transactional(self, get_user_mock, braze_mock, _, __):
```
stephendherrera: To be precise: get_user_mock is actually bound to send_tx_messages, and braze_mock to get_user_data. The real braze mock is the unused _ param, so the assertions never touch it.

basket/admin.py
```diff
@@ -171,50 +143,34 @@ def dsar_unsub_view(self, request):
...
 
-                def handler(emails, use_braze_backend=False):
+                def handler(emails):
                     # Process the emails.
                     for email in emails:
-                        if use_braze_backend:
-                            contact = braze.get(email=email)
-                        else:
-                            contact = ctms.get(email=email)
+                        contact = braze.get(email=email)
```
stephendherrera: braze.get()/braze.delete() (line 204) can also raise BrazeNotConfigured, which isn't caught here. get_user_data in utils.py, which handles it cleanly. Will 500 instead if Braze isn't configured.

stephendherrera (changes requested):
Added detail on three of copilots findings and flagged one it missed:

- models.py (kwarg retry): affects QueuedTask.retry() too, not just FailedTask, it doesn't kwarg filter at all.
- utils.py (HTTPError handling): confirmed dead code, since Braze already converts HTTP errors into its own exception types before they reach this except clause. Needs to catch those instead of requests.exceptions.HTTPError.
- test_upsert_user.py (mock ordering): confirmed the mocks are mislabeled, braze_mock is actually bound to get_user_data, so the test isn't checking the right thing. Does that make sense? 
- admin.py (new): dsar_unsub_view/dsar_delete_view don't catch BrazeNotConfigured the way get_user_data does, so they'll 500 instead of failing cleanly.

I'd say the utils.py and models.py retry issues are blockers pre-merge since they're live on every request/retry.

stephendherrera (approved):
This looks like it'll work great!

PR #1774, opened 2026-09-15
Create new basic contact endpoint

maribedran:
Create a new "basic" contact endpoint without some of the fields from the enterprise contact endpoint, and include a separate field for the terms and conditions (opt-in is already used to accept communications). The data is sent to the same Google Sheets file.


basket/contact/api.py
```diff
@@ -6,42 +6,42 @@
...
+from .schemas import ContactBasicSchema, ContactEnterpriseSchema
 
 ### /api/v1/contact URLS
 contact_router = Router()
 
 
+def submit_contact(request, payload, form: str):
+    """Rate limit, drop bots, then queue the contact. Shared by every contact form."""
+    for reason, key in (("ip", "ip"), ("email", lambda *_: payload.business_email.lower())):
+        if is_ratelimited(request, group=f"basket.contact.{form}.{reason}", key=key, rate=settings.CONTACT_ENTERPRISE_RATE_LIMIT, increment=True):
```
stephendherrera, line 18: This now applies `CONTACT_ENTERPRISE_RATE_LIMIT` to both the basic and enterprise groups even though the group name is already parameterized by form. Should we consider different limits? 

We could be ok, but wanted to surface...

basket/contact/backends/google_sheets_contact_sink.py
```diff
@@ -22,23 +44,7 @@ def submit(self, contact: dict) -> None:
...
-            contact["opt_in"],
-            contact["lead_source"],
-            contact["cta"],
-            contact["firefox_use_stage"],
-            contact["deployment_size"],
-            contact["support_needs"],
-            contact["timeline"],
-            contact["message"],
-        ]
+        row = [contact.get(column, "") for column in _COLUMNS]
```
stephendherrera, line 46: This column mapping is going to throw off Google Sheets. Was `accepted_terms` explicitly requested as a submission log item? There are quite a few app script logic nodes in the sheet that are tied to index values, so untangling will create quite a complex issue.

basket/contact/api.py
maribedran: No preference. Up to you.

basket/contact/backends/google_sheets_contact_sink.py
maribedran: The ticket asked for a separate privacy checkbox, but we don't actually need to send it to the sheet. We only need the opt-in there. I'll remove this one from the sheet and keep it only in the schema so we reject a submission without explicit acceptance.

basket/contact/api.py
stephendherrera: Let's try sharing and we can adjust if needed.

stephendherrera (approved):
LGTM

